PR #45, opened 2015-11-01
Documentation setup - install statinamic as dependencies

magsout:
If I follow your instructions  here http://moox.io/statinamic/docs/setup/#install-from-npm

You recommand to install `statinamic` as a `devDependencies`. If I follow this, all `dependencies` are added as `devDependencies `in my package.json. 

After that, I have to run the following: `statinamic setup`  But this time all `dependencies` are copy as `dependencies`, so I have this kind of error each time I run `npm install`:

```console
npm WARN package.json Dependency 'postcss-loader' exists in both dependencies and devDependencies, using 'postcss-loader@^0.7.0' from dependencies
```
So I think it's better to provide in your instruction to install `statinamic` as `dependencies`

### DIFF
--- a/demo/content/docs/setup.md
+++ b/demo/content/docs/setup.md
@@ -6,7 +6,7 @@ title: How to setup statinamic
 
 ```console
 $ npm install
-$ npm install --save-dev statinamic
+$ npm install --save statinamic
 ```
 
 Statinamic require some boilerplate, in order to provide you some flexibility.


MoOx:
Hum... Let me think. 

> You recommand to install statinamic as a devDependencies. If I follow this, all dependencies are added as devDependenciesin my package.json. 

What are the "all dependencies" your are mentioning ?

You should not have some duplicates. But I guess that npm are installing peerDeps automatically and save them in the same "(dev)deps" section you install your first package? Note: that will not be the case anymore in npm3, which is by default with node 5.

Indeed, `statinamic setup` install statinamic required peer deps as deps, I made this to simplify, because in a "private" project (statinamic is not likely to be part of another package, but who knows...), it can make sense to put in deps only deps that will be used on the client side, and put in dev deps, the deps only used to build your thing.

Anyway, that's clear that we should have all the things in one place. I guess devDeps make more sense than deps? Not sure. Deps or dev deps for statinamic and its (peer)deps ?

magsout:
>What are the "all dependencies" your are mentioning ?

my `package.json: https://gist.github.com/magsout/5fc177916c8892bd3d6c

```console
npm install --save-dev statinamic
```

my `package.json`: https://gist.github.com/magsout/d704ecc61cdca5caabf9

```console
npm run statinamic
```

my `package.json`: https://gist.github.com/magsout/819e1ccc6c076656f8e6

MoOx:
Now the command save peer deps as dev deps.

magsout:
:+1:

PR #50, opened 2015-11-07
Add Window suppport

MoOx:
This PR is just to double check that the deploy will not happen (since travis build PR 2 times, from the origin and from a merged branch called "master").


iamstarkov:
do you know you can white-list and black-list branches on which travis will run builds?

MoOx:
I want travis to run tests on all branches, but only deploy on master. That's ok, travis handle that correctly https://github.com/MoOx/statinamic/blob/master/.travis.yml#L6-L10

PR #54, opened 2016-01-03
Update .gitignore

mxstbr:
`.DS_Store` is a [useless file to have in a .git repo](https://en.wikipedia.org/wiki/.DS_Store) but every Mac will add it to the folders.

I added it to the `.gitignore` so nobody gets another persons `.DS_Store`.

EDIT: Added the npm error file, `npm-debug.log` to the `.gitignore` as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,6 @@ docs/scripts/statinamic-static.js
 
 # some tests output some files
 src/md-collection-loader/__tests__/output/
+
+# Default file on Mac
+.DS_Store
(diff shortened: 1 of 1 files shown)


MoOx:
I like to keep this kind of crap out of repos by adding that in my global gitignore https://github.com/MoOx/setup/blob/master/dotfiles/gitignore_global
But maybe that make the repo more accessible?

mxstbr:
I wanted to push a PR, and accidentally included both those files. Not everybody has a global gitignore, i prefer it in the repo-specific one.

MoOx:
Fair enough.

mxstbr:
:+1:

PR #55, opened 2016-01-03
Update to Babel 6

mxstbr:
WIP

For some reason the `npm run test` fails – any idea why?

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,3 @@
+{
+  "presets": ["es2015", "stage-0", "react"]
+}
--- a/package.json
+++ b/package.json
@@ -50,9 +50,13 @@
     "webpack-nano-logs": "^1.0.0"
   },
   "devDependencies": {
-    "babel": "^5.8.23",
-    "babel-core": "^5.8.23",
+    "babel": "^6.0.0",
+    "babel-cli": "6.3.17",
+    "babel-core": "^6.0.0",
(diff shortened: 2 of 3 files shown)


MoOx:
It will be a bit more complicated since we have to handle peerdeps and also deps of the docs/ part that use the parent folder as a node_modules (statinamic use itself for its doc)

mxstbr:
Can you point me in a direction of what to do?

MoOx:
You need to update peerDependencies versions too, and also update the docs/package.json.
And please try to not commit changes for the eslint config.
Also I would like to keep babel config in package.json to avoid the .*rc hell :)

mxstbr:
Alright, will take another look in the afternoon! 

Didn't even notice the eslint config changes, any idea how they happened? I didn't change it...

MoOx:
Each time you run an npm command that adjust the package.json, this section is updated. I will outsource this config as a eslint shareable package to avoid that because it's a pain in the arse :)

mxstbr:
:+1: airbnb/javascript maybe?

iamstarkov:
👍 for airbnb/javascript

MoOx:
I am on it since I am currently upgrading lot of things (tape>ava which use Babel 6...) in order to be able to have a code coverage percentage :)

mxstbr:
:+1:

PR #89, opened 2016-01-22
Consolidate import statement

thangngoc89:
First easy PR


MoOx:
Why did you close?

thangngoc89:
@MoOx You use this kind of import everywhere. So I think I should open another PR which fix all of them?

Or is this your recommend ?

MoOx:
Haha, maybe a new PR then ?
Thanks for this.

PR #90, opened 2016-01-23
Ignore statinamic-static.js file

thangngoc89:
instead of scripts folder

### DIFF
--- a/boilerplate/scripts/.gitignore
+++ b/boilerplate/scripts/.gitignore
@@ -1,0 +1,1 @@
+statinamic-static.js


thangngoc89:
It looks like master branch is different from my local version. It'll take a look later

MoOx:
Not sure for this change, this file is the server side script and should not be considered as output as it is the script used to generate a part of the output (static html files). Maybe a "build" folder?

thangngoc89:
Yup. Any folder but scripts folder (or I can just add a .gitignore for that file)

MoOx:
There is already a line to ignore it https://github.com/MoOx/statinamic/blob/master/boilerplate/.gitignore#L3
Not working for you?

thangngoc89:
Oops. Sorry, I use my own .gitignore file.
Should be close

Issue #60, opened 2016-01-03
Upgrade redux & redux devtools


MoOx:

thangngoc89:
@MoOx close this ?

MoOx:
Closed by f7552492b19d6a503cc8c15e16863c9a991f6ec2

PR #91, opened 2016-01-23
[WIP] Upgrade redux

thangngoc89:
Ref: #60


src/client/index.js
```diff
@@ -45,10 +37,12 @@ export default function statinamic({
     <div id="statinamic-container">
       <MetadataProvider metadata={ metadata }>
         <Provider store={ store }>
-          <Router history={ history } routes={ routes } />
+          <div>
```
MoOx: Not a fan about the extra div.
FYI, the server side is here: https://github.com/MoOx/statinamic/blob/cb000b78d0172e05dadffeb8bdb0fecbe770/src/static/to-html/url-as-html.js

MoOx:
Can you rebase? Master should be good now. Also please consider my note about the div.

src/client/index.js
thangngoc89: @MoOx 
How would you avoid extra div ? 

devtools component needs to be inside Provider
and Provider only accepts a child

MoOx: Oh yes, indeed. so can you adjust the static generation part?

thangngoc89: okie. Pushed some code.
waiting for Travis

thangngoc89:
@MoOx rebased, ready to be merged

MoOx:
Thanks!

Issue #96, opened 2016-01-24
collection pages are not correctly prerendered (or the opposite)


MoOx:
Ref https://github.com/MoOx/statinamic/issues/92

MoOx:
Ref https://github.com/MoOx/statinamic/issues/92

thangngoc89:
In my case, that is the opposite

thangngoc89:
@MoOx is this ok now?

MoOx:
Will double check on a big project when upgrade will be done. But I hope this will be good :)

thangngoc89:
I hope so.  I dont have time to upgrade my blog atm

thangngoc89:
I confirm that this issue was fixed in 0.7.0

MoOx:
Cool.

Issue #92, opened 2016-01-24
Bundle collection in window.__INITIAL_STATE__


thangngoc89:
I followed your instruction on create a collection page [here](http://moox.io/statinamic/docs/usage/collections/)

[Here is my collection source code](https://github.com/thangngoc89/blog/blob/master/web_modules%2Fcontainers%2FArchive%2FArchive.js)

But React will throw a warning:

```
Warning: React attempted to reuse markup in a container but the checksum was invalid. This generally means that you are using server rendering and the markup generated on the server was not what the client was expecting. React injected new markup to compensate which works but you have lost many of the benefits of server rendering. Instead, figure out why the markup being generated is different on the client or server:
 (client) MgeFRiHrav1TDm192y99 _3jzDdN3QzHnR0O_efk
 (server) MgeFRiHrav1TDm192y99" href="/" data-reac
```

Demo: http://khoanguyen.me (open your console)

After poking around your code, I see the issue is from this line 
https://github.com/MoOx/statinamic/blob/master/src%2Fstatic%2Fto-html%2Furl-as-html.js#L75

Is it posible that we could pass an option to bundle `collection` in `INITIAL_STATE` on specific page ?

MoOx:
I think we have two issues here, so I am going to open a new one for the bug, and keep this one as a feature request.

I already thought about embeding collections directly, but for now I use statinamic on a website with hundreds of posts, so I was thinking that could be better to not include collection in all pages, but well, since user is supposed to only call ONE html pages per session, should be embed.

(The real issue is that rendering collections from json and from node cache does not give the same thing, which lead to a react warning: https://github.com/MoOx/statinamic/issues/96)

thangngoc89:
> The real issue is that rendering collections from json and from node cache does not give the same thing, which lead to a react warning

No, there is no different between collections from json and from node cache.
The initial state of `collections` in client is a blank array `[]` 

On a slow connection, you could see the splash between `full collection ---> no entry ---> full collection`

MoOx:
Oh right. So I will change this. You are right.

PR #98, opened 2016-01-25
Add option to bundle collection into html

thangngoc89:
This PR add an option to bundle collection to html

ref #92 #96 

Usage:

```
// index.md
---
layout: Homepage
title: Home page
collection: true
---
```

I'm replying on Travis for test right now, because ava freezes my laptop

### DIFF
--- a/src/PageContainer/__tests__/component.js
+++ b/src/PageContainer/__tests__/component.js
@@ -42,8 +42,8 @@ test("should render a Page if page is ok", () => {
 test(
   "should render a visible error if page is not ok and no PageError available",
   () => {
-  const renderer = createRenderer()
-  renderer.render(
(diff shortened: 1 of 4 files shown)


MoOx:
I think this should be default behavior. I don't want a weird "collection" option in meta (and you don't want neither, am I right?).
Maybe a "collection: false" to disable in frontmatter, and a collection: true by default in js config?

Also you will need to adjust the useless fetch for the collection.json file if the collection is already in the initial state.

MoOx:
Did you try "ava src/ a single file.js" to run test?

thangngoc89:
Oh yeah. you are right. this should be the default behavior.

thangngoc89:
Oops. Accidentally delete this branch. Continue as #100

Issue #23, opened 2015-09-22
Dev server should generate static version for url on the fly


MoOx:
Currently we need to build a static version + run the dev server. This is counter productive. The express server should be able to generate the page on the fly directly.

MoOx:
@thangngoc89 just to let you know that I am on this, and this will clean a LOT OF THINGS :)

thangngoc89:
Ok.  Good to know . I means time. I'll just blogging then

MoOx:
ETA: Almost ready :)

thangngoc89:
Can't wait ....

MoOx:
OMG it works. I finally fixed [the last issue I got](https://github.com/istarkov/babel-plugin-webpack-loaders/issues/13#issuecomment-180278051) (bad usage of babel-plugin-webpack-loaders).
I still have to cleanup things now and will push my work as a PR so you can review it (warning: huge pr) ^^

thangngoc89:
Finally :)

thangngoc89:
Awesome. About that config, webpack 2 have a better syntax to define loader with query. no more ugly JSON.stringify

MoOx:
Yeah I see that.

thangngoc89:
Why don't use NODE_ENV for easier babel config ?

There is 

- [cross-env](https://www.npmjs.com/package/cross-env) for windows ENV 
- [better-npm-run](https://www.npmjs.com/package/better-npm-run) for a cleaner run script in package.json

MoOx:
Because NODE_ENV=static is used for babel-plugin-webpack-loaders and the client webpack build is run in the same process. babel-plugin-webpack-loaders does not make sense for client side code so I must disable it for client side script.

thangngoc89:
Ah. That makes sense.

PR #100, opened 2016-01-25
Bundle collection to html by default

thangngoc89:
Continue from #98 
ref #92 #96 

This PR will bundle collection to html by  default.

We can disable this behavior on page specific by passing an option `collection: false` to it's head

Example:
```
---
layout: Post
title: Some page you don't want to bundle collection by default
collection: false
---
```

TODO:

- [x] Add collection to bundle
- [x] Don't fetch collection when it is already in store
- [x] Allow to disable this behaviour
- [ ] Fix tests

### DIFF
--- a/src/client/index.js
+++ b/src/client/index.js
@@ -6,6 +6,10 @@ import { Provider } from "react-redux"
 
 import fetchJSON from "../fetchJSON"
 import MetadataProvider from "../MetadataProvider"
+import {
+  set as collectionSet,
+  error as collectionError,
+} from "../redux/modules/collection"
 
 export default function statinamic({
   metadata,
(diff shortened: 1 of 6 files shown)


thangngoc89:
@MoOx could you help me with these test ?
I'm not so familiar with ava.

MoOx:
Can you rebase/squash on master and drop the changes in src/PageContainer/__tests__/component.js? (I fixed this)

statinamic
```diff
@@ -0,0 +1 @@
+node_modules/statinamic
```
MoOx: what is this file?

MoOx:
For ava you should be able to debug using `ava --verbose the/file/to/run.js`.

MoOx:
FYI, I think for now, I will just completely remove collection.json, so we can even drop this option.
I don't really see any use case for having a json if we already have the collection in any entry points.
Make sense?

statinamic
thangngoc89: oops. my bad

thangngoc89:
Yes. Since we only fetch collection.json once. And we did bundle it inside every HTML file. We can just remove it and make code cleaner

MoOx:
I am on it. This require serving prerendered pages during development (which is not the case atm, see #23)

thangngoc89:
Rebased

thangngoc89:
@MoOx should I remove the option to not bundle collection in HTML ?

MoOx:
Probably. I think we should not touch this PR until I end my current refactoring. I will notify you when it's done.

thangngoc89:
Close in favor of #113

MoOx:
Note that the PR #113 description contained enough ("close #113") to automatically close this issue when merged into master :)

thangngoc89:
Didn't know about that :+1:

PR #101, opened 2016-01-26
Add travis cache for node_modules

thangngoc89:
This could speed up build a little speed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,10 @@ node_js:
   - '4'
   - '5'
 
+cache:
+  directories:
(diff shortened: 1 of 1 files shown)


MoOx:
Yeah but you might miss some minor updates/patches that might introduce regressions. Not a fan :/

thangngoc89:
Okie. It's up to you

MoOx:
I think it's better to way maybe a little longer, but that might help us catch regressions due to deps changes more quickly!

MoOx:
Thanks!

Issue #93, opened 2016-01-24
Pass custom store enhancer and middleware to redux store


thangngoc89:
Some basic function that I could build when statinamic have this feature:

- Search : https://github.com/treasure-data/redux-search

I could send a PR for this. But I'm waiting for #91

MoOx:
Two things: 

- I was thinking about removing redux because I am not sure it's really useful https://github.com/MoOx/statinamic/issues/40, feel free to give your thoughts about this
- if we keep it, have you any idea about a nice flexible way to allow that?

thangngoc89:
> have you any idea about a nice flexible way to allow that

Allow 2 extra args in createStore function:

A proof of concept

```js
// https://github.com/MoOx/statinamic/blob/master/src/redux/createStore/index.js#L4
import { createStore, applyMiddleware, compose } from "redux"
import thunk from "redux-thunk"

export default function(reducer = {}, initialState = {}, extraMiddlewares = {}, storeEnhancers = {}) {
  function promiseMiddleware() {
    return (next) => (action) => {
      const { promise, types, ...rest } = action
      if (!promise) {
        return next(action)
      }
      else if (!promise.then) {
        throw new Error(
          "promiseMiddleware expects a promise object that implements then()"
        )
      }

      const [ REQUEST, SUCCESS, FAILURE ] = types
      next({ ...rest, type: REQUEST })
      return promise.then(
        (response) => next({ ...rest, response, type: SUCCESS }),
        (error) => next({ ...rest, error, type: FAILURE })
      )
    }
  }

  let finalCreateStore

  if (__DEVTOOLS__) {
    const { devTools, persistState } = require("redux-devtools")

    finalCreateStore = compose(
      applyMiddleware(promiseMiddleware, thunk, ...extraMiddlewares),
      devTools(),
      persistState(window.location.href.match(/[?&]debug_session=([^&]+)\b/)),
      ...storeEnhancers
    )(createStore)
  }
  else {
    finalCreateStore = applyMiddleware(promiseMiddleware, thunk)(createStore)
  }

  return finalCreateStore(reducer, initialState)
}

```js

MoOx:
Well, why not.

thangngoc89:
Close https://github.com/MoOx/statinamic/commit/ba169122851bfd30c5dfa4f8f0fbe1a41aed6145

PR #102, opened 2016-01-26
Extra middlewares and store enhancers for redux store

thangngoc89:
ref #93


MoOx:
Can you also add something in changelog and maybe in docs/content/docs/advanced/redux.md as well?

thangngoc89:
Added changelog and docs

MoOx:
:+1:

PR #106, opened 2016-01-26
Fix build script

thangngoc89:
FYI, you are serving raw script in docs site

### DIFF
--- a/boilerplate/scripts/build.js
+++ b/boilerplate/scripts/build.js
@@ -166,7 +166,7 @@ builder({
         { disable: config.dev }
       ),
 
-      ...config.prod && [
+      ...config.production && [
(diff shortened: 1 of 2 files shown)


MoOx:
I will review this because some others flag use prod and should use production. (eg: `__PROD__`).

thangngoc89:
Well. Or just add -- prod in build script 

https://github.com/MoOx/statinamic/blob/master/docs/package.json#L55
becomes `babel-node scripts/build --static --production --prod`

That's faster

MoOx:
--production --prod... Hum.
production or prod? One should be enough.

thangngoc89:
Haha. That's why I file this PR

thangngoc89:
Close in favor of #113

Issue #11, opened 2015-09-09
🎉 hot loading for markdown


MoOx:
- use js instead of json
- use md to react instead of raw html https://www.npmjs.com/package/commonmark-react-renderer

thangngoc89:
Any progress ?

MoOx:
I will handle this after #23, #92 and #92. Then I will cut a release.
If you have idea(s), let's discuss about it :)

thangngoc89:
Well. Maybe we should let webpack handle those file loading for us.  statinamic only need to keep track of a mapping between filename and url

This is a proofs of concept : 
webpack will generate N + 1 chunk (1 chunk per markdown files)

```
export default function loadAsync(filename, callback) {
  require([ "bundle!../content/" + filename + ".md" ], bundledResult => {
    bundledResult(result => callback(result))
  })
}
```

MoOx:
It won't be that easy. Currently data are loaded via from json via a fetch() request. We need to rething a huge part. See https://github.com/MoOx/statinamic/pull/107#issuecomment-175195475

thangngoc89:
Gatsby uses json https://github.com/gatsbyjs/gatsby/blob/2b0fffa0669d47f2798478f5dd22cd511cc83187/lib/loaders/markdown-loader/index.js#L38 and dangerouslySetInnerHTML https://github.com/gatsbyjs/gatsby/blob/2b0fffa0669d47f2798478f5dd22cd511cc83187/lib/isomorphic/wrappers/md.jsx#L17 but still get hot loading

MoOx:
gatsby includes ALL content in one entry, this way, it's "easy".

thangngoc89:
@MoOx yeah. That's why I deleted my comment.

MoOx:
Haha ok :)

thangngoc89:
My approach in #107 was using webpack code spliting (one markdown file per chunk)

But when it came to hot reload, I'm stuck

MoOx:
The problem is that one route (PageContainer) require all md. I will think about that when #23 is fixed.

thangngoc89:
Just wait for #23 to be fixed. I think I'll close #109. I mind have a better idea

thangngoc89:
With #151 ,  now we have a map between real url and source file. I think I'll give this another try

MoOx:
This should be pretty easy to handle in PageContainer using module.hot (just retrigger getPage).

thangngoc89:
[...]

Issue #97, opened 2016-01-24
Keep "layouts" reference in React context, not in Redux store


thangngoc89:
because functions can't be serialized to JSON. This breaks persistState in devTools

MoOx:
Any exact pointers on what is causing an issue for you?

thangngoc89:
I'm testing on #91 

First, save it: http://localhost:3000/?debug_session=foo
Open new tab, go to that address

`pageComponents={}` 

Basicaly, persistState dump the current state to localStorage. Since functions can't be serialized to JSON, It would break persisState functionality

MoOx:
Oh I see. Yeah I think we should use react context for this. Don't you think?

thangngoc89:
Yup. Please read my comment at #40

MoOx:
So will move "layouts" (I just renamed "pageComponents" as "layouts" in master) out of the store, and will use context for this.

thangngoc89:
:+1:

MoOx:
That's already good in master https://github.com/MoOx/statinamic/blob/813aafc3b32608a7281270610905f4000e95f133/src/ContextProvider/index.js

PR #107, opened 2016-01-26
proof of concept for load data via Webpack

thangngoc89:
ref #11
There is no hot loading yet. Will figure it out

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
     "fs-extra": "^0.26.4",
     "js-beautify": "^1.5.10",
     "lnfs-cli": "^1.0.1",
+    "lodash": "^4.0.1",
     "markdown-it": "^5.0.0",
     "npmpub": "^3.0.0",
     "nyc": "^5.2.0",
--- a/src/redux/modules/index.js
+++ b/src/redux/modules/index.js
@@ -1,3 +1,4 @@
 export collection from "./collection"
 export layouts from "./layouts"
 export pages from "./pages"
+export loadPage from "./loadPage"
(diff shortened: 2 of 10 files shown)


package.json
```diff
@@ -71,6 +71,7 @@
     "fs-extra": "^0.26.4",
     "js-beautify": "^1.5.10",
     "lnfs-cli": "^1.0.1",
+    "lodash": "^4.0.1",
```
MoOx: I really don't want to include lodash. We can do 90% of lodash with simple es6.

MoOx:
I don't think it's possible to enhance the current codebase. This have to be rewritten. And the fact that we use dangerousinnerhtml make this "stupid" for hot loading.
Currently, json files are already emmited by webpack.

I don't really understand what this solution brings us.

The have hot loading we will have to use some `if (module.hot)` blocks to reload data when webpack emit a .hot.js file.

I think it will not be that easy.

src/redux/modules/index.js
```diff
@@ -1,3 +1,4 @@
 export collection from "./collection"
 export layouts from "./layouts"
 export pages from "./pages"
+export loadPage from "./loadPage"
```
MoOx: the loading to load page should be in "pages" module.

package.json
thangngoc89: This is just a proof of concept when I finish. ok ?

src/redux/modules/index.js
thangngoc89: loadPage should be in context like `layouts`

#97

package.json
MoOx: ok :)

thangngoc89:
Close as this is not the way to get this done :)

PR #109, opened 2016-01-27
Hot loading for markdown files

thangngoc89:
out

This time, I took a different approach.

In dev, just bundle everything into a big file using `require.context` and enable module.hot with it
In prod, use webpack bundle-loader.

TODO: 

- [ ] Move getPage(loadAsync) function some where not redux store (context ? )
- [x] Remove lodash
- [ ] Remove all debug code added
- [ ] Remove index.json file (handle these LOC https://github.com/MoOx/statinamic/blob/master/src/builder/index.js#L64-L70)
- [ ] Remove fetch 
- [ ] Figure out how to handle error case on chunk loading
- [ ] Move build folder somewhere (or just delete it after build)

### DIFF
--- a/docs/scripts/load-async.js
+++ b/docs/scripts/load-async.js
@@ -1,0 +1,29 @@
+import enhanceCollection from "statinamic/lib/enhance-collection"
+
+const getFileFromContext = (context, filename) => {
+  const bundledResult = context.keys().map(context)
+  const result = enhanceCollection(bundledResult, {
+    filter: { __filename: filename },
+    limit: 1,
+  })
+
+  return result[0]
+}
+
+export default function loadAsync(filename, callback) {
+  if (__PROD__) {
+    const bundledResult = require("bundle!../content/" + filename)
+    bundledResult(result => callback(result))
+  }
+  else if (__DEV__) {
+    let context = require.context("../content", true, /\.md$/)
+    callback(getFileFromContext(context, filename))
+
+    if (module.hot) {
+      module.hot.accept(context.id, function() {
+        context = require.context("../content", true, /\.md$/)
+        callback(getFileFromContext(context, filename))
+      })
+    }
+  }
+}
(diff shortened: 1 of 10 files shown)


docs/scripts/load-async.js
```diff
@@ -0,0 +1,29 @@
...
+    filter: { __filename: filename },
+    limit: 1,
+  })
+
+  return result[0]
+}
+
+export default function loadAsync(filename, callback) {
+  if (__PROD__) {
+    const bundledResult = require("bundle!../content/" + filename)
```
MoOx: The problem is here. You will have to lazyload file.
You are not supposed to use dynamic require because webpack will bundle every thing in that folder in the bundle.
I think the require(bundle should be in the loader itself, but I am not sure, I am still using my brain to handle server side pre-rendering for dev (to be able to catch checksums issues).

MoOx:
The hard part will be to make the code ok for production, with lazy loading.

thangngoc89:
> The problem is here. You will have to lazyload file.

Any pointer?  Currently, webpack generate N+1 chunk with each markdown file per chunk.  When navigate to different route,  it automatically downloads the correct chunk.  But throw the above error. Which is pretty weird to me.

thangngoc89:
> The hard part will be to make the code ok for production, with lazy loading.

I think it's impossible with current architecture.  Refactor needed

MoOx:
Yeah. I think we should wait a bit that I finish the dev server pre-rendering (this one makes me rething everything about the current process, which was kind of a ugly hack).

thangngoc89:
Well. I figure the bug out 

```
  target: "node",
```

Those chunk was build using staticConfig. That explain the global `exports` stuff. I think this PR hits a block. waiting for your refactor those

thangngoc89:
@MoOx pushed new commit. It's working now on production build. Is this the right approach for hot loading ?

thangngoc89:
Did you try this branch ? @MoOx

Good enough?  Or we need a different approach

thangngoc89:
This would allow long-term caching since we let webpack handle everything

thangngoc89:
There is a lot of changes in master since this pr. Rebasing will be hard. Close this for now

Issue #69, opened 2016-01-06
Highlight code correctly in doc


MoOx:
For now it's pretty shitty.

mxstbr:
Have done dynamic client-side code highlighting with prism.js for http://sharingbuttons.io, can take a look if you want me to. 

A solution that adds the highlighting on build would probably be nicer though...

MoOx:
Currently the code is processed during the "markdown to html" conversion so it's prerendered.
I am using highlight.js but with a stupid custom theme. Maybe just doing CSS will fix the thing. I will see.

thangngoc89:
An example https://github.com/thangngoc89/blog/blob/1c16d9063c5f86b3f54e2ffeaa39e3b39157b4bd/web_modules/styles/vendors/_hljs.scss

You can get different theme from 
https://highlightjs.org/static/demo/

(just view source code for css)

MoOx:
Thanks, will check this and update doc too :)

thangngoc89:
I filed an PR on #110

MoOx:
Closed by 23a781ef24ccd44dd291d7d2e6f1ee874bdf1ceb

PR #110, opened 2016-01-28
Add hightlight.js styling for docs

thangngoc89:
ref #69 
screenshot from 2016-01-28 18 49 59


MoOx:
Thanks.
I think we should put this in a separate css file, included "I don't no where but not in the current file" :D
Maybe a highlightjs.css file included in an entry point?

thangngoc89:
Pushed new commit. Hightlight.js style is in it's own file

MoOx:
Thanks, that's better with these colors! I will probably tweak those in the future, but better than it was.

PR #112, opened 2016-02-06
Move opinionated deps to optionalPeerDependencies

### LINKED ISSUE #99: Move some opinionated peerDependencies as optionalDeps (or somewhere else)
IMHO, I would remove these packages as peerDependencies. They are too opinionated

```
"eslint-config-i-am-meticulous": "^2.0.0",
```

Maybe others eslint packages as well. User can add there own preset of eslint or using a different linting options

The same reason for stylelint

```
"stylelint": "^4.0.0",
"stylelint-config-standard": "^1.0.0",
```

What if I want to use SASS instead of cssnext ?

```
"postcss-browser-reporter": "^0.4.0",
"postcss-cssnext": "^2.4.0",
"postcss-loader": "^0.7.0",
"postcss-reporter": "^1.3.0",
```

thangngoc89:
This PR close #99


package.json
```diff
@@ -89,45 +89,47 @@
...
-    "eslint-config-i-am-meticulous": "^2.0.0",
-    "eslint-loader": "^1.1.0",
-    "eslint-plugin-react": "^3.6.2",
     "extract-text-webpack-plugin": "^0.8.2",
     "file-loader": "^0.8.1",
     "highlight.js": "^8.5.0",
     "invariant": "^2.1.1",
     "json-loader": "^0.5.2",
     "markdown-it": "^5.0.0",
-    "markdown-it-toc-and-anchor": "^1.0.1",
```
MoOx: Can you also move markdown-it?
We should accept any parser (eg: mdast).

src/bin/statinamic-setup.js
```diff
@@ -1,12 +1,17 @@
 import npmInstallPackage from "npm-install-package"
 
-import { peerDependencies as peerDeps } from "../../package.json"
+import {
+  peerDependencies as peerDeps,
+  optionalPeerDependencies as opPeerDeps,
+} from "../../package.json"
+
+const toBeInstalledDeps = Object.assign({}, peerDeps, opPeerDeps)
```
MoOx: Can you use spread operator instead of Object.assign?

`const toBeInstalledDeps = { ...peerDeps, ...opPeerDeps }`

MoOx:
You can rebase and maybe also add most webpack loaders?

src/bin/statinamic-setup.js
thangngoc89: Will do

thangngoc89:
I think I'll start from scratch.  Tired of rebase

package.json
thangngoc89: :+1: valid point. Anyone should be able to use there own markdown loader

MoOx:
What is the problem with a rebase?

thangngoc89:
Conflict.  Don't worry,  i'll force update this branch.

thangngoc89:
@MoOx could you have a look ?

MoOx:
lgtm. Thanks!

PR #113, opened 2016-02-06
Avoid using a webpack builded file for static build

### LINKED ISSUE #23: Dev server should generate static version for url on the fly

### LINKED ISSUE #92: Bundle collection in window.__INITIAL_STATE__

MoOx:
By using webpack on the fly, we can now pre-render page during
development. And static build is faster.

Will close #23, close #92, close #100, close #106 

I stil have 2 things to fix (some tests broken + css file link in static html)

@thangngoc89 you can already take a look and try if you want :)


thangngoc89:
:+1

Omg.  It's huge

thangngoc89:
@MoOx there is some bug that prevent react bootstrap on client side:

```
Warning: React.createElement: type should not be null, undefined, boolean, or number. It should be a string (for DOM elements) or a ReactClass (for composite components).warning @ (program):45
0.0.0.0/:46 Uncaught Invariant Violation: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined.
```

Repro:

- Fresh clone
- npm install
- cd docs && npm install
- npm i -D raw-loader react-svg-inline
- npm install ../
- npm start

src/client/index.js
```diff
@@ -2,43 +2,34 @@
...
-        error,
-      })
-    )
-
   ReactDOM.render(
     <div id="statinamic-container">
-      <MetadataProvider metadata={ metadata }>
-        <Provider store={ store }>
+      <StatinamicContextProvider
+        collection={ typeof window !== undefined && window.__COLLECTION__ }
```
thangngoc89: `typeof window !== 'undefined'` ?

src/client/index.js
```diff
@@ -2,43 +2,34 @@
 import React from "react"
 import ReactDOM from "react-dom"
 import Router, { browserHistory } from "react-router"
```
thangngoc89: This should be `{ Router, browserHistory } from ...`

src/client/index.js
```diff
@@ -2,43 +2,34 @@
...
 
 export default function statinamic({
+  layouts,
   metadata,
   routes,
   store,
 }) {
   let devtools = false
-  if (__DEVTOOLS__) {
+  if (process.env.REDUX_DEVTOOLS) {
```
thangngoc89: There should be a check for window object since devtools require window

MoOx: The idea is to avoid this code when builded in production. A test on window. will make this build to have redux devtools code all the time and we don't want that. That's why I used process.env.* which can be replaced by webpack with DefinePlugin. That's said for now I only inject NODE_ENV, so I will add this one and document this.

MoOx:
I have probably forget some deps since static build "works for me" :)
Will double check and update PR.

thangngoc89:
`npm prune` to remove unlisted deps

src/client/index.js
thangngoc89: This line https://github.com/MoOx/statinamic/blob/babel-node-loaders/src/redux/createStore/index.js#L33 should be 

```
process.env.REDUX_DEVTOOLS && 
typeof window !== "undefined"
```

because on line 38 of the same file. We are using window object which is undefined in static build.

Anyway, just run `REDUX_DEVTOOLS=true npm start` and you will see a bunch of error

MoOx: The thing is: we need a test that can be statically converted when uglifying. I will add a process.env.CLIENT then...

thangngoc89: :+1:

MoOx:
npm pruned, branch rebased, commit updated/amended.

Use "npm test" instead of your combo (npm test use a symlink trick to make the things works properly, like in "real word", when statinamic is at the same level as other deps).

thangngoc89:
You forgot to remove emitting collection.json in md-collection-loader

MoOx:
Good catch.
I will remove it.
I think I will also add some "funtional" testing using the docs/dist folder. It's not easy to handle a full coverage for this kind of project :)
I have to go for today and will handle that this evening or tomorrow.

thangngoc89:
Thank you for all of your hard works :+1: . Can't wait until this PR get merged

thangngoc89:
No HMR for React components ?

MoOx:
It seems that I have introduced a bug. I will review that.

thangngoc89:
You removed babel-preset-react-hmre from source code

MoOx:
I will add it back in boilerplate. That's because I add to remove it from global config.

MoOx:
I just removed collection.json and fixed hmre.

MoOx:
For npm I don't have anything magic. npm@3 is slow. Did you try latest 3.7.1 which is supposed to be faster?

MoOx:
The fuck babel. This is very annoying. The issue "Cannot read property 'start' of undefined" is still here on another part of the code...

thangngoc89:
Lock it to 6.4.5 ?

MoOx:
Not a solution. Ok this bug just have been fixed in babel-plugin-webpack-loaders.

thangngoc89:
Ha. That's why it's faliing on others project of mine

MoOx:
Just use `"babel-plugin-webpack-loaders": "^0.2.2"`.

thangngoc89:
This may be not related to this PR, but can you add some sourcemap ? It makes debugging much better

MoOx:
No idea how to add that :)

thangngoc89:
https://webpack.github.io/docs/configuration.html#devtool

tl;dr

Add this to webpack config on development:

`devtool: "cheap-module-eval-source-map"`

MoOx:
Feel free to make a PR for devtools.

PR #119, opened 2016-02-10
Don't include assets config by default

thangngoc89:
If there is no `content/assets` fs-extra will throw an error about no file/directory


thangngoc89:
It looks like docs and boilerplate is replying on this default behaviour. @MoOx could you confirm

MoOx:
The default boilerplate have some assets, so I think we found another solution.
Maybe only include it if the folder exist and if not show a clear warning?

thangngoc89:
Great.

PR #125, opened 2016-02-11
[WIP] Read assets file name (js, css) from webpack stats instead of hardcoded name

thangngoc89:
This would allow user to use a real CDN for long-term caching.

There is 1 issue those, I had to hardcoded file name in dev mode 

https://github.com/MoOx/statinamic/pull/125/files#diff-9ec6170964a6a4d01ea02104c51b3d84R65

Looking for a way to read it from wepackCompiler

### DIFF
--- a/docs/scripts/webpack.config.babel.js
+++ b/docs/scripts/webpack.config.babel.js
@@ -87,7 +87,7 @@ export default {
   ),
 
   plugins: [
-    new ExtractTextPlugin("[name].css", { disable: config.dev }),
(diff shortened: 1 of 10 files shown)


MoOx:
I will take it and try to get the filename from the compiler instance (otherwise will open an issue). Thanks!

thangngoc89:
since webpackCompiler is a async action. I don't think we can find a way to read it in dev mode.

MoOx:
I just did. 7f1e8cd4c6f078872b0bb402b53f1a8d3e435a21

thangngoc89:
Awesome :+1: :+1:

Issue #117, opened 2016-02-10
npm doesn't resolve fs-extra even though it's in devDeps


thangngoc89:
Well. On a fresh start of boilerplate, I get this error : 

```js
> @ start /home/khoa/web/postcss.org
> cross-env BABEL_ENV=statinamic babel-node scripts/build --server --open --dev

/home/khoa/web/postcss.org/node_modules/statinamic/lib/builder/index.js:21
  _fsExtra2.default.emptyDirSync(destination);
                    ^

TypeError: _fsExtra2.default.emptyDirSync is not a function
    at exports.default (/home/khoa/web/postcss.org/node_modules/statinamic/lib/builder/index.js:21:21)
    at Object.<anonymous> (build.js:12:1)
    at Module._compile (module.js:398:26)
    at loader (/home/khoa/web/postcss.org/node_modules/babel-register/lib/node.js:134:5)
    at Object.require.extensions.(anonymous function) [as .js] (/home/khoa/web/postcss.org/node_modules/babel-register/lib/node.js:144:7)
    at Module.load (module.js:344:32)
    at Function.Module._load (module.js:301:12)
    at Function.Module.runMain (module.js:430:10)
    at /home/khoa/web/postcss.org/node_modules/babel-cli/lib/_babel-node.js:161:27
    at Object.<anonymous> (/home/khoa/web/postcss.org/node_modules/babel-cli/lib/_babel-node.js:162:7)

```

```shell
✔ ~/web/postcss.org [statinamic-0.7 L|✚ 2] 
22:56 $ node -v
v5.3.0
✔ ~/web/postcss.org [statinamic-0.7 L|✚ 2] 
22:56 $ npm -v
3.7.1
```

No idea why npm doesn't resolve fs-extra

MoOx:
Can you try "npm install" again? Just to be sure?
Sometimes I have some deps that are missing that are downloaded after a 2nd npm install... (yeah that sucks).

thangngoc89:
Nope. Run `npm install` again didn't help

MoOx:
I don't want to add a dep for "we don't know why it happens". We need to identify the source of the issue.

thangngoc89:
Alright. Changed the title.

thangngoc89:
This might be an issue with babel (again).

MoOx:
GREAT.

MoOx:
how do you explain this works on travis?
Did you try a full clean npm install one more time, just to be sure?

[...]

PR #126, opened 2016-02-12
Remove mkdirp in favor of fs-extra

thangngoc89:
Changed:

- Remove mkdirp and rimraf in source
- Remove mkdirp in deps list (can't remove rimraf because it is used in cli too)
- Move fs-extra to deps instead of devDeps (hopefully this will solve #117)


MoOx:
Nice.

PR #128, opened 2016-02-12
Allow no assets config

thangngoc89:
Since `undefined` is invalid in json and we have a default assets config. So `config.assets !== undefined` always true

### DIFF
--- a/src/builder/index.js
+++ b/src/builder/index.js
@@ -26,7 +26,7 @@ export default function(options) {
 
   if (config.static) {
     // Copy static assets to build folder
-    if (config.assets !== undefined) {
+    if (config.assets) {
       const copyDest = path.join(destination, config.assets.route)
       fs.copySync(config.assets.path, copyDest)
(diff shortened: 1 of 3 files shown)


MoOx:
Good catch!

MoOx:
Maybe we could even just do `if (config.assets) {` to allow `false`?

thangngoc89:
Yup. That's valid. I update this pr

thangngoc89:
@MoOx now user can use false, null to disable assets option

MoOx:
nice!

thangngoc89:
@MoOx I forgot about minimist. assets is now accepting `string`, `null`, `boolean`. Should we just remove minimist validation for `assets` ?

MoOx:
Yeah sure please!

MoOx:
@thangngoc89 I am fixing that covered by tests.

thangngoc89:
:+1: great

PR #132, opened 2016-02-13
Changing nodejs_version order on Appveyor

thangngoc89:
Since npm@3 is a lot faster then npm@2. Testing node 5 first should give us faster respond about failing in window

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,8 +2,8 @@
 
 environment:
   matrix:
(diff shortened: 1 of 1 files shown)


MoOx:
Good idea. Btw, any idea why it's broken on windows? Maybe related to the difference of the way "symlink" works on windows?! It was working before the babel-plugin-... refactoring.

PR #135, opened 2016-02-13
Allow to disable auto open new tab when starting dev server

### LINKED ISSUE #134: Disable auto open new tab when starting dev server
It looks like there is no way to overwrite this 

https://github.com/MoOx/statinamic/blob/master/src%2Fbuilder%2Fserver.js#L21

Why? It just too annoying for me

thangngoc89:
This will close #134 
Could you cut a release after merging this ? You know. It just too annoying.

### DIFF
--- a/src/configurator/index.js
+++ b/src/configurator/index.js
@@ -13,6 +13,7 @@ const defaultOptions = {
   devHost: "0.0.0.0",
   devPort: "3000",
   verbose: false,
+  open: true,
 }
 
 const defaultOptionsCLIonly = {
@@ -35,6 +36,7 @@ const fieldTypes = {
   "production": "boolean",
   "static": "boolean",
   "server": "boolean",
+  "open": "boolean",
 }
 
 export default function config(pkg = {}, argv = process.argv) {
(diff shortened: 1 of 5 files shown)


src/configurator/index.js
```diff
@@ -20,6 +20,7 @@ const defaultOptionsCLIonly = {
   production: false,
   static: false,
   server: false,
+  open: false,
```
MoOx: should be "true" by default. Otherwise it's a breaking change, not a fix :)

MoOx: Also maybe this option should not be "cli only" so we can adjust from package.json

thangngoc89: Alright, so this should be disabled via package.json. but can also be enabled via cli

thangngoc89:
`open` option should be able to disabled via package.json. but can also be able to enabled via cli.

MoOx:
Can you add it in the Configuration doc section? Ad use "Added:" not "Fixed:" in CHANGELOG.

thangngoc89:
Should be good now

MoOx:
Will merge when merging "next" branch (this PR scope is master).

thangngoc89:
:+1:

MoOx:
Conflict so I cherry picked/updated the commit directly in master.

PR #138, opened 2016-02-13
Next version 0.7.0+

MoOx:
(To avoid online doc being outdated, that might be a solution (work on a "next" branch where we can submit PR on (PRception)).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: node_js
 
 node_js:
-  - '4'
   - '5'
+  - '4'
 
 matrix:
   fast_finish: true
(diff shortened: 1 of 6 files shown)


MoOx:
Will release after my bednight :)

thangngoc89:
we need to talk about workflow a little bit. I'm tracking `next` in some branchs. Maybe just don't auto deploy docs on Travis and do it locally when releasing ?

MoOx:
I think keeping master as a reflect of the current released package is a good thing. And when public API become stable enough, having doc auto deployed is also a good thing.
Maybe I should just not delete the "next" branch. I will keep it up to date with master.

PR #145, opened 2016-02-15
Clean node cache when running dev server

thangngoc89:
- Combine layouts, metadata, routes into data object
- Some minor change in web_modules/layouts/index.js

~~Don't merge yet. I messed up something here. The original branch works like a charm~~

### DIFF
--- a/boilerplate/scripts/build.js
+++ b/boilerplate/scripts/build.js
@@ -1,8 +1,6 @@
 import builder from "statinamic/lib/builder"
 
-import * as layouts from "../web_modules/layouts"
-import metadata from "../web_modules/app/metadata"
-import routes from "../web_modules/app/routes"
+import * as exports from "../web_modules/app/exports"
 import store from "../web_modules/app/store"
 
 import config from "./config.js"
@@ -14,8 +12,6 @@ builder({
   webpackConfig,
   clientWebpackConfig,
 
-  layouts,
-  metadata,
-  routes,
+  exports,
   store,
 })
--- a/boilerplate/web_modules/app/exports.js
+++ b/boilerplate/web_modules/app/exports.js
@@ -1,0 +1,3 @@
+export const layouts = require.resolve("../layouts")
+export const metadata = require.resolve("./metadata")
+export const routes = require.resolve("./routes")
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+- Changed: build and client scripts now require an exports value
+  that must contains layouts, metadata and routes
+  ([#145](https://github.com/MoOx/statinamic/pull/145))
+- Added: during development for pre-rendering, dev server will refresh
+  all files for each render
+  ([#145](https://github.com/MoOx/statinamic/pull/145))
+
 # 0.7.1 - 2016-02-12
 
 _Minor release since the only change only affects new setup (so this is not
--- a/docs/scripts/build.js
+++ b/docs/scripts/build.js
@@ -1,8 +1,6 @@
 import builder from "statinamic/lib/builder"
 
-import * as layouts from "../web_modules/layouts"
-import metadata from "../web_modules/app/metadata"
-import routes from "../web_modules/app/routes"
+import * as exports from "../web_modules/app/exports"
 import store from "../web_modules/app/store"
 
 import config from "./config.js"
@@ -14,8 +12,6 @@ builder({
   webpackConfig,
   clientWebpackConfig,
 
-  layouts,
-  metadata,
-  routes,
+  exports,
   store,
 })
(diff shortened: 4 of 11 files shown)


boilerplate/web_modules/layouts/index.js
```diff
@@ -1,5 +1,12 @@
-export Page from "./Page" // default layout
-export PageError from "./PageError"
+import Page from "./Page" // default layout
+import PageError from "./PageError"
 
 // you can add your own layout here
-// eg: export Post from "Post"
+// eg: import Post from "Post"
+
+const layouts = {
```
MoOx: I don't really like this breaking change (not necessary). Maybe in the dev server we should do something smarter instead of just .default?

boilerplate/scripts/build.js
```diff
@@ -1,8 +1,8 @@
 import builder from "statinamic/lib/builder"
 
-import * as layouts from "../web_modules/layouts"
-import metadata from "../web_modules/app/metadata"
-import routes from "../web_modules/app/routes"
+const layouts = require.resolve("../web_modules/layouts")
+const metadata = require.resolve("../web_modules/app/metadata")
+const routes = require.resolve("../web_modules/app/routes")
```
MoOx: Maybe we can move this 3 lines in an "web_modules/app/exports.js" and reuse it in the index-client.js ?

boilerplate/scripts/build.js
```diff
@@ -1,8 +1,8 @@
...
 import store from "../web_modules/app/store"
```
MoOx: Should we introduce the store too?

boilerplate/scripts/build.js
```diff
@@ -14,8 +14,10 @@ builder({
   webpackConfig,
   clientWebpackConfig,
 
-  layouts,
-  metadata,
-  routes,
   store,
+  data: {
```
MoOx: I don't like the "data" name. Maybe "exports" ?

boilerplate/web_modules/layouts/index.js
thangngoc89: Alright, let's make it smarter :))

boilerplate/scripts/build.js
thangngoc89: Alright. Will rename it

boilerplate/scripts/build.js
thangngoc89: Nope. The store must be in memory and hold data. If we include it, this will become a big refactor

MoOx: The store only keep page data. Collection (during development) are in another memory cache.

thangngoc89:
Forgot about collection. React will complaint about invalid checksum when we change the markdown content. Btw, renamed data to exports, this PR should be backward compat with 0.7.1

boilerplate/web_modules/app/exports.js
```diff
@@ -0,0 +1,9 @@
+const layouts = require.resolve("../layouts")
+const metadata = require.resolve("./metadata")
+const routes = require.resolve("./routes")
+
+export default {
+  layouts,
+  metadata,
+  routes,
+}
```
MoOx: Maybe we can make the code smaller by directly exporting values?

boilerplate/scripts/build.js
thangngoc89: Introduce the store to exports var just add more complexity, no real benefit.

MoOx:
As soon as someone need to update it script/build or something in this folder, it's actually a breaking change.

thangngoc89:
Before statinamic reach 1.0, public API will keep changing...

MoOx:
Yeah but that not just an addition :) (no big deal, just to be clear).

MoOx:
There is in this PR a real change (even if it's really minor)

thangngoc89:
So change CHANGELOG from 'Added' to 'Changed' right ?

MoOx:
You can probably add two entries? 

- Changed: build and client scripts now require an ``exports`` value that must contains ``layouts``, ``metadata`` and ``routes``
- Added: during development for pre-rendering, dev server will refresh all files for each render

thangngoc89:
Updated changelog.

MoOx:
Nice. Will merge and release with custom routes feature when I have some times.

thangngoc89:
Squashed. Just let you know that client script wasn't changed in this PR.

thangngoc89:
@MoOx could you merge this into next branch ? I'm gonna send a different request that has conflict with this one

MoOx:
Yeah but maybe we should change this in order to "normalize" build and client bootstraps?

thangngoc89:
I'm on it

MoOx:
Ok I will see that change for client script later.

PR #151, opened 2016-02-16
Custom routes

MoOx:
No more "random" generated/consumed URL.
md-collection-loader now add fields in addition to __url: __resourceUrl (actual file, so might contains /index/html) and __dataUrl (json file url).

.md files accept an "route" field for custom urls.

Consequence: PageContainer now know before fetching if a url is know or unknow.


thangngoc89:
Bravo :clap:

src/builder/index.js
```diff
@@ -6,7 +6,7 @@ import debug from "debug"
 import webpack from "./webpack"
 import devServer from "./server"
 
-import filenameToUrl from "../filename-to-url"
+import collection from "../md-collection-loader/cache"
```
thangngoc89: This is in memory cache. Will it cause invalid checksum in dev server?

MoOx: Should not :) And if it create some, we will fix it somehow :)

thangngoc89: Ha. It's not

src/builder/server.js
```diff
@@ -119,29 +123,22 @@ export default (webpackConfig, options = {}) => {
...
+        !req.originalUrl.endsWith("/")
+      ) {
         res.redirect(req.originalUrl + "/")
       }
 
+      const relativeUri = item.__dataUrl.replace(config.baseUrl.pathname, "")
+      const filepath = join(config.cwd, config.destination, relativeUri)
+      const fileContent = memoryFs.readFileSync(filepath)
       log(
         `Using '${ filepath }' to pre-render '${ req.originalUrl }' (${ uri })`
```
thangngoc89: I think this log line is useless now.

src/builder/index.js
MoOx: ;)

src/builder/index.js
```diff
@@ -33,24 +33,16 @@ export default function(options) {
...
 
-      // There is probably a better way to get markdown as json without reading
-      // fs, but I am tired
-      const pagesData = {}
+      // TODO use a more reliable way to only get entry points
       const assetsFiles = {
         css: [],
         js: [],
       }
       const assets = stats.compilation.assets
```
thangngoc89: Is `stats.compilation.namedChunks` better ?

src/builder/index.js
```diff
@@ -33,24 +33,16 @@ export default function(options) {
...
-        if (name.endsWith("index.json")) {
-          const url = filenameToUrl(name)
-          pagesData[url] = JSON.parse(assets[name]._value)
-        }
-
         if (name.endsWith(".js")) {
           assetsFiles.js.push(name)
         }
-
         if (name.endsWith(".css")) {
```
thangngoc89: Also I think this can be `else if` no need for another condition (my fault)

MoOx: yeah, no big deal :)

MoOx:
I will fix windows (I used path.join() for url, which is incorrect, so I will create a small util to handle that since there is nothing native in node).

thangngoc89:
`path.sep()` ? It will return `\` in windows and `/` in linux

MoOx:
That what is using path.join(), but url are the same on windows and linux haha.
So I will push something like


```js
const urlJoin = () => (path.join(...args).replace(path.sep, "/"))
```

What do you think?

thangngoc89:
Should replace it everywhere

`const re = new RegExp(path.sep, g)`

thangngoc89:
If I'm offline and navigate to a valid page, a 404 page will be shown. Is this an intended behaviour ? Can we provide a offline mode page ?

MoOx:
If you have it in page memory you won't have a 404. If you do not have it in memory, this make sense to show a 404 (like any other website/app will). (Btw, same behavior as before this PR).

By offline mode you mean keep page in localstorage? If so we will need to invalidate the cache, that will probably be more work :) But why not.

thangngoc89:
I think the behaviour before this PR is it'll show nothing except an uncatch promise error in console. Showing an 404 error when fetch fail will lead to misunderstood

thangngoc89:
About offline mode,  it should be in another PR imho

MoOx:
If you fetch (via xhr or via a normal link), you will get a 404 anyway. The 404 came from the network error. The only place the 404 are hardcoded is if you call a link that is not know in the collection or if you hit the 404.html

thangngoc89:
Oh right. Got it :)

MoOx:
https://github.com/MoOx/statinamic/blob/82b5ca6b50f0d0247e78befa9520829a8480f4c8/src/fetchJSON/index.js#L20

Issue #88, opened 2016-01-19
Lint markdown in boilerplate with remark-lint


MoOx:

thangngoc89:
Some tools:

- https://github.com/DavidAnson/markdownlint
- https://github.com/wooorm/remark-lint (cli avaliable)

thangngoc89:
This should be easy now since we are using remark as default markdown engine

MoOx:
Boilerplate/base theme will be gone in 1.0 and the example will be minimalistic.

PR #158, opened 2016-02-18
Generate meta description and bundle it to collection

### LINKED ISSUE #79: Add something to easily generate a meta description from page content
Talking about these LoC in the boilerplate: https://github.com/MoOx/statinamic/blob/3ba0b9ffd691f48b6554359bd6f73f04d70af684/boilerplate/web_modules/Page/index.js#L6-L8

What is the use of having that function?

thangngoc89:
This PR will close #79 

This is base on https://github.com/simbo/metalsmith-better-excerpts (MIT license)

Features:

- Add a description option to md-collection-loader
- Default options:

```
const defaultOpts = {
   pruneLength: 140,
   pruneString: "…",
 }
```

- Description generation order : 

  + Read it from front matter --> return untouch
  + Get x letters and generate prune (get the hold word, don't cut in the middle)


src/md-collection-loader/index.js
```diff
@@ -100,6 +96,10 @@ module.exports = function(input) {
     ...metadata,
   }
 
+  if (query.description) {
+    mdObject = description(mdObject, query.description)
+  }
+
```
MoOx: Don't think let is necessary, this make the code not clear. Btw I am not sure we should rely and some html output. This makes things complicated.

MoOx:
What do you think about not using html rendered output as a base and use markdown to text instead? (or something like that)?

src/md-collection-loader/index.js
thangngoc89: Alright. Remove it

thangngoc89:
@MoOx I'm open to that. But I don't really know how to convert markdown into plain text . Any suggest?

MoOx:
I am looking for a solution right now :D

thangngoc89:
Maybe this one https://github.com/stiang/remove-markdown ?

MoOx:
Maybe using remark with a plugin (even a custom/stupid one that strips nodes)

thangngoc89:
Also replying on a generated HTML output is not a good idea. It will break test if the renderer changed

MoOx:
You can try this one yes.

thangngoc89:
I'll use this package instead https://github.com/wooorm/strip-markdown since it's from @-wooorm (mdast author)

MoOx:
Ok sure :)

thangngoc89:
@MoOx please take a look. Updated the first post.

I also pull in remark (required for strip-markdown). We can use it to lint markdown (#88)

src/md-collection-loader/__tests__/description.js
```diff
@@ -0,0 +1,83 @@
...
+
+test("should generate description with default option", (t) => {
+  const mdObject = {
+    head: {},
+    rawBody: fixtures.basic,
+  }
+
+  const expectedResult = "Lorem ipsum\n\nDolor sit amet, consectetur adipisicing elit, sed do eiusmod\ntempor incididunt ut labore et dolore magna aliqua. Ut enim ad…"
+
+  t.same(description(mdObject).head.description, expectedResult)
```
MoOx: You can use "is" here (=== for strings).

src/md-collection-loader/__tests__/description.js
```diff
@@ -0,0 +1,83 @@
...
+  const mdObject = {
+    head: {
+      description: "Tokyo rain bridge motion fetishism boat dissident long-chain hydrocarbons disposable sprawl warehouse cardboard Kowloon neon face forwards claymore mine spook cartel voodoo god film tattoo girl dome realism.",
+    },
+    rawBody: "",
+  }
+
+  const expectedResult = "Tokyo rain bridge motion fetishism boat dissident long-chain hydrocarbons disposable sprawl warehouse cardboard Kowloon neon face forwards claymore mine spook cartel voodoo god film tattoo girl dome realism."
+
+  t.same(description(mdObject).head.description, expectedResult)
```
MoOx: t.is too

MoOx: And to make the test more straight forward, please expect the same value (mdObject.head.description).

src/md-collection-loader/__tests__/description.js
```diff
@@ -0,0 +1,83 @@
...
+    rawBody: fixtures.basic,
+  }
+
+  const opts = {
+    pruneLength: 50,
+    pruneString: "[...]",
+  }
+
+  const expectedResult = "Lorem ipsum\n\nDolor sit amet, consectetur[...]"
+  t.same(description(mdObject, opts).head.description, expectedResult)
```
MoOx: t.is

src/md-collection-loader/__tests__/description.js
thangngoc89: Sorry. I'm new to ava

src/md-collection-loader/__tests__/description.js
thangngoc89: :+1:

src/md-collection-loader/description.js
```diff
@@ -0,0 +1,39 @@
+import remark from "remark"
+import stripMd from "strip-markdown"
+import { prune }  from "underscore.string"
+
+const defaultOpts = {
+  pruneLength: 140,
+  pruneString: "…",
```
MoOx: Is the pruneString considered when counting and splitting value?

"Limited to X chars" => "Limited to X char…"  (this is a stupid cut :D)

thangngoc89: pruneString was not count into length
Test: https://github.com/epeli/underscore.string/blob/master/tests/prune.js

"Limited to X chars"

prune 6 '...'

"Limited..." (always get the hold word)

src/md-collection-loader/description.js
```diff
@@ -0,0 +1,39 @@
...
+}
+
+export default function description(mdObject, opts = {}) {
+  opts = { ...defaultOpts, ...opts }
+
+  if (opts.pruneLength <= 0) {
+    throw new Error("Prune length must be larger than 0")
```
MoOx: Instead of throwing an error, I would prefer a warning more explicit like

```js
console.warning(
`You defined 'description.pruneLength' of md-collection-loader with an zero value. ` +
`This does not make sense, so the default value ${ defaultOpts.pruneLength } has been used. `
)
// or something like that.


// then use default value instead of 0.
// And maybe we can show this warning and use default when value is < 10 ?
```

I like having (and will try to do this more often) really verbose error/warning like React does.
I think it's really helpful.

thangngoc89: Will do. I'm really like verbose error message

CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+- Added: Generate meta description and bundle it to collection
```
MoOx: Can you add a pointer for people that will upgrade? Like adding:

```
See ``boilerplate/web_modules/layouts/Page/index.js`` for an example.
```

Where do you think we should add doc for this?

thangngoc89: I think it should be in webpack configuration. ref #160

MoOx:
Great work. Almost ready to merge ;)

src/md-collection-loader/description.js
thangngoc89: How to mock console.warning ? I usually use a sinon for this job.

thangngoc89:
@MoOx should be ready now. Do you think mocking console.warn is neccessary ?

CHANGELOG.md
MoOx: Yeah.

MoOx:
mocking console, not a necessity imo. Reading stdout maybe? Or what about adding an argument like `console = console` in fn definition (so we can override console for test).

thangngoc89:
I did that. See https://github.com/thangngoc89/statinamic/commit/1d4843e318bd3925f8d0ec29b13ef5dcfa218a7b

thangngoc89:
You should let me squash them. They're a mess.

MoOx:
Who cares. Before 1.0 :D
Anyway, maybe try to amend/force push all the time in a PR so PR is always ready to be merged?

MoOx:
I will rewrite next branch history, don't worry :)

thangngoc89:
I had bad experience with it. :-1: . Keeping commits would help reviewers easy to keep track of PR

MoOx:
Ok so I will ask before merging a release, or I will make a script that will squash and merge :)

thangngoc89:
The second option is even better. Would be awesome if Github have that "button"

PR #162, opened 2016-02-18
Combine routes, metadata, layouts into exports var on client side

thangngoc89:
As suggested in https://github.com/MoOx/statinamic/pull/145#issuecomment-184228487 , we already doing this in server side (#145)

### DIFF
--- a/boilerplate/scripts/build.js
+++ b/boilerplate/scripts/build.js
@@ -1,6 +1,6 @@
 import builder from "statinamic/lib/builder"
 
-import * as exports from "../web_modules/app/exports"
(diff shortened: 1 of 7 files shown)


MoOx:
Not a fan of this one. We are supposed to remove code, not add more.

thangngoc89:
Can't think of a better way. Webpack will complaint if using `require(require.resolve(...))`

MoOx:
I am just going to revert the thing I asked you in the first place (I will drop exports.js file).

thangngoc89:
Your call. Feel free to close this PR

PR #163, opened 2016-02-18
Allow to start static server serving build dist

thangngoc89:
I often use python simpleHTTPServer to server static build dist. But as `statinamic docs` is not in domain's root, I had to do something hacky like this

```shell
npm run build && \
mv dist statinamic && \
mkdir dist && \
mv statinamic dist/ && \
start-server
```

With this PR, I can simply doing this `npm run build -- --server`

or better, add a npm script like `npm run build:server`

### DIFF
--- a/src/builder/index.js
+++ b/src/builder/index.js
@@ -62,6 +62,17 @@ export default function(options) {
         exports,
         store,
       })
+      .then(() => {
+        if (config.server) {
+          devServer(null, { config })
+        }
+      })
+      .catch((error) => {
+        log(color.red("✗ Faild to start static server"))
+        setTimeout(() => {
+          throw error
+        }, 1)
+      })
     })
   }
   else if (config.server) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
   ([#104](https://github.com/MoOx/statinamic/pull/104))
 - Fixed: ability to pass custom webpack devServer config
   ([#157](https://github.com/MoOx/statinamic/issues/157))
+- Added: cli option to server static dist build `npm run build -- --server`
+  ([#163](https://github.com/MoOx/statinamic/pull/163))
 - Added: during development for pre-rendering, dev server will refresh
   all files for each render
   ([#145](https://github.com/MoOx/statinamic/pull/145))
--- a/src/builder/server.js
+++ b/src/builder/server.js
@@ -31,151 +31,162 @@ export default (webpackConfig, options = {}) => {
     )
   }
 
-  const devEntries = [
-    require.resolve(`webpack-hot-middleware/client`),
-  ]
-
-  const devConfig = {
-    ...webpackConfig,
-    // debug: true,
-    // watch: true,
-    // colors: true,
-    entry: {
-      // add devEntries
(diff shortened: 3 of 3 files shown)


MoOx:
What is the point? What do you want to serve from the build (which is for development)?

thangngoc89:
Sometime, devServer complaint about invalid checksum. (like upgrade css-module)

I want to server this to make sure this isn't happen on static build

MoOx:
Can't you "just" reuse the code of the existing server, without starting the webpack step/middleware?

thangngoc89:
Done. I thought that it would just make devServer more complicated

thangngoc89:
Github diff in this situation is awful

Here is a more clearer diff

```
const server = new express()

if (config.static) {
// serve static dist here
}
else {
// old devServer
}

server.listen
```

thangngoc89:
Oops. Sorry Travis https://travis-ci.org/MoOx/statinamic/jobs/110059266#L2088

MoOx:
Haha. Just killed the process.

src/builder/index.js
```diff
@@ -62,16 +62,19 @@ export default function(options) {
...
+  if (config.server && !config.static) {
     devServer(options.clientWebpackConfig, {
       config,
       exports,
       store,
     })
   }
-  else {
+
+  if (!config.server && !config.static) {
```
MoOx: maybe you can make the all file more readable by using if/else if/else ?

thangngoc89: Done. Please let's me know if everything is ok and I'll squash these commits

src/builder/index.js
```diff
@@ -62,9 +62,10 @@ export default function(options) {
         exports,
         store,
       })
+      .then(() => devServer(null, { config }))
```
MoOx: Not sure it's ok here. You launch the devServer for every static build?

src/builder/index.js
```diff
@@ -62,9 +62,10 @@ export default function(options) {
...
     })
   }
-  else if (config.server) {
+  else if (config.server && !config.static) {
```
MoOx: since it's in the 'else' you don't need to test config.static.

src/builder/index.js
thangngoc89: Yup. It's not ok at all. Added catch block to catch this error

MoOx:
Nice. Maybe we should document this?

thangngoc89:
Could you suggest where to add it ? Or maybe even better. Add a `npm run build:server` command ?

thangngoc89:
Squashed all commits

MoOx:
Not sure this will be used that much (it should not on the long run). You know what, we are not going to document this for now :)

thangngoc89:
:+1: looks like I'm the only one who's using this

MoOx:
Haha. We will see. I hope that a temporary wait to "debug" :/

thangngoc89:
I hope so.

MoOx:
Can you rename your commit message (eg: just copy the changelog line?)

thangngoc89:
Done. Didn't meant to force push when rebasing

thangngoc89:
Thanks you. This makes working on #165 easier

PR #170, opened 2016-02-21
Add source map for better DX

thangngoc89:
ref https://webpack.github.io/docs/configuration.html#devtool

### DIFF
--- a/boilerplate/scripts/webpack.config.babel.js
+++ b/boilerplate/scripts/webpack.config.babel.js
@@ -6,6 +6,9 @@ import pkg from "../package.json"
 import config from "./config.js"
 
 export default {
+  ...config.dev && {
+    devtool: 'cheap-module-eval-source-map',
+  },
   module: {
     loaders: [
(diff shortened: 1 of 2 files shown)


boilerplate/scripts/webpack.config.babel.js
```diff
@@ -6,6 +6,9 @@ import pkg from "../package.json"
 import config from "./config.js"
 
 export default {
+  ...config.dev && {
+    devtool: 'cheap-module-eval-source-map',
```
MoOx: Any point why this value and not another?
webpack devtool values are really confusing.
Btw, it's weird that the linter doesn't warn for singles quotes...

thangngoc89: Ah. Because linter can auto fix this. 

This value is recommended by my team leader. It just work for me.

MoOx: Ok doesn't warn cause --fix :D

PR #172, opened 2016-02-21
build: Reliable way to get assets' name

thangngoc89:
Sort assets A-Z for predictable chunks order

### DIFF
--- a/src/builder/index.js
+++ b/src/builder/index.js
@@ -33,20 +33,28 @@ export default function(options) {
     webpack(options.clientWebpackConfig, log, (stats) => {
       log(color.green("✓ Static assets: client build completed"))
 
-      // TODO use a more reliable way to only get entry points
       const assetsFiles = {
         css: [],
         js: [],
       }
-      const assets = stats.compilation.assets
-      Object.keys(assets).forEach((name) => {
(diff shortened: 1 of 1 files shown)


MoOx:
What is the benefit of this as a end user? Not sure what does this brings.

thangngoc89:
You meant the sorting or the way to get assets filename ?

thangngoc89:
For the sorting, webpack always put `manifest` chunk after main chunk. (see long term caching https://github.com/MoOx/statinamic/issues/130#issuecomment-184660624)

thangngoc89:
Getting assetsByChunkName in webpack docs https://github.com/webpack/docs/wiki/long-term-caching#get-filenames-from-stats

MoOx:
Maybe we should manually "sort"? Because is someone add a "Aaaa.js" this will be before manifest.

thangngoc89:
So put manifest as the first element? I think it add more stuff to maintaince. Just tell people that chunks are sorted a-z, named your chunk to get the right order.

MoOx:
I will think about that one more time. I doesn't looks like a intuitive experience...

MoOx:
I guess it's better than nothing.

PR #176, opened 2016-02-22
Intercerpt internal link to avoid fullpage reload

thangngoc89:
This is a working progress. Looks like I need to re-write the whole library to handle React lifecycle hooks

### DIFF
--- a/src/PageContainer/component.js
+++ b/src/PageContainer/component.js
@@ -9,6 +9,14 @@ const isDevelopment = process.env.NODE_ENV !== "production"
 const isClient = typeof window !== "undefined"
 const isDevelopmentClient = isDevelopment && isClient
 
+let catchLinks
+let browserHistory
+
+if (isClient) {
+  catchLinks = require("../_utils/catch-links").default
+  browserHistory = require("../client").browserHistory
+}
+
 export default class PageContainer extends Component {
 
   static propTypes = {
@@ -35,8 +43,26 @@ export default class PageContainer extends Component {
     this.preparePage(this.props, this.context)
   }
 
+  componentDidMount() {
+    this.catchInternalLink()
+  }
+
   componentWillReceiveProps(nextProps) {
     this.preparePage(nextProps, this.context)
+  }
+
+  componentDidUpdate() {
+    this.catchInternalLink()
+  }
+
+  catchInternalLink() {
+    if (!isClient) {
+      return
+    }
+    catchLinks(this._content, (href) => {
+      const pathname = href.replace(process.env.STATINAMIC_PATHNAME, "")
+      browserHistory.push(pathname)
+    })
   }
 
   preparePage(props, context) {
@@ -114,7 +140,7 @@ export default class PageContainer extends Component {
     const Layout = this.getLayout(this.props, this.context, page)
 
     return (
-      <div>
+      <div ref={ (ref) => this._content = ref }>
         {
           !page.error && page.loading && PageLoading &&
           <PageLoading />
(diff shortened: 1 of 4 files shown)


src/PageContainer/component.js
```diff
@@ -35,10 +43,24 @@ export default class PageContainer extends Component {
     this.preparePage(this.props, this.context)
   }
 
+  componentDidMount() {
+    this.catchInternalLink()
+  }
```
MoOx: You should also do that in componentDidUpdate too.

thangngoc89: I know. Any idea how to removeEventListener ?

src/PageContainer/component.js
```diff
@@ -114,7 +136,7 @@ export default class PageContainer extends Component {
     const Layout = this.getLayout(this.props, this.context, page)
 
     return (
-      <div>
+      <div ref="content">
```
MoOx: Don't use string, it's deprecated. Use `ref={ (c) => this._content = c`.

src/PageContainer/component.js
MoOx: https://facebook.github.io/react/docs/component-specs.html tells me that this should be in componentWillUnmount. But not sure if that make sense to stop listening since removed (dropped) from the dom. The GC should handle this right?

thangngoc89: I think so. Just pushed some commits

src/PageContainer/component.js
```diff
@@ -35,8 +43,27 @@ export default class PageContainer extends Component {
     this.preparePage(this.props, this.context)
   }
 
+  componentDidMount() {
+    this.catchInternalLink()
+  }
+
   componentWillReceiveProps(nextProps) {
     this.preparePage(nextProps, this.context)
+    console.log(this._content)
```
MoOx: console.log here :)

src/PageContainer/component.js
```diff
@@ -35,8 +43,27 @@ export default class PageContainer extends Component {
...
+
+  componentDidUpdate() {
+    this.catchInternalLink()
+  }
+
+  catchInternalLink() {
+    if (!isClient) {
+      return
+    }
+    catchLinks(this.refs.content, (href) => {
```
MoOx: this.refs.content > this._content

thangngoc89:
Should be good by now

thangngoc89:
Fixed tests and rebased.

MoOx:
Can you find a way to cover this with test somehow? I would like avoiding decreasing the percentage of coverage from now on (to reach 100%).

thangngoc89:
We need a DOM to test what I added.

MoOx:
Are you afraid of jsdom ? :D

thangngoc89:
Never use jsdom before. Most of my projects use Karma with Chrome or nothing

MoOx:
using jsdom is very easy compared to requiring a browser or karma: https://github.com/tmpvar/jsdom#easymode-jsdomenv

MoOx:
I will try to add test for this without a dom by just adding some mock.

thangngoc89:
Thanks. Anything before 0.8.0 ?

MoOx:
I will review docs and cut the release soon.

MoOx:
Maybe I will add more tests to reach 70 or 80%.

thangngoc89:
I'll see what I can do to help. I'm writing some common cases for invalid checksum. Point to our Link wrapper. But English is harder than Javascript :))

PR #177, opened 2016-02-22
 Pump dependencies

thangngoc89:
There should be a way for greenkeeper to pump these peerDep and optionalDep for us.


MoOx:
I think it should be "Bump dependencies" :)
Thanks!

Issue #153, opened 2016-02-16
Smart Offline mode via Service Worker.


MoOx:
- Keep pages data in localstorage
- invalidate the cache via a manifest when site is updated.

Ref https://github.com/MoOx/statinamic/pull/151#issuecomment-184595116

thangngoc89:
This may help https://github.com/NekR/offline-plugin

thangngoc89:
Here is the plan:

- [x] Appcache
- [ ] Service worker
- [ ] Don't download a data file twice. Keep it in localStorage.

MoOx:
service worker and/or app cache are supposed to handle the download right?

thangngoc89:
With service worker, we can control how browser download file. Appcache only kicks in whem offline

MoOx:
It seems that the "offline" plugin might helps indeed. Did you take a look to it?

thangngoc89:
I tried offline plugin. Not useful in ourcase since it only can index assets emited from webpack

MoOx:
We emit json via webpack, so are you sure that's not relevant?

thangngoc89:
We have assets file that will be copied with fs, index.html file generated by `url-to-html` , and "offline plugin" doesn't have a fallback option for service worker

MoOx:
Maybe assets in the assets folder should be "optional" (cached when downloaded?)
For the index.html we can probably make an exception.
Just asking that to see if we can integrate serviceworker easily.

thangngoc89:
Why all the troubles ? Just write a service worker logic and put it in.

MoOx:
"Just write", ok then :)

thangngoc89:
fyi, I'm working on service worker. A json file will never be downloaded twice.

thangngoc89:
>  A json file will never be downloaded twice

Haha. Appcache is doing this for us. JSON files was loaded from cache already. No don't see the benefits of writing SW anymore. 

Compare browsers compatiable : [appcache](http://caniuse.com/#search=appcache) and [sw](http://caniuse.com/#search=serviceworker)

MoOx:
appcache is brutal and cache "all the things". sw is a smarter method that will allow you to cache only things already request. For big sites, appcache make no sense (too long to cache everything) and sw will make much more sense.

thangngoc89:
The simplest sw tutorial ever.
https://remysharp.com/2016/03/22/the-copy--paste-guide-to-your-first-service-worker

[...]

PR #190, opened 2016-02-24
Added: generate appcache manifest for offline access

thangngoc89:
This is a working progress for offline mode (#153)

- Only the `/index.html` will be cached and will be used as fallback for other endpoints
- You can specify your own [globby](https://github.com/sindresorhus/globby) patterns (it's all yours, be careful)


thangngoc89:
Ah. I forgot, should I specify a FALLBACK section ? 

If I specify the /index.html, there will be a redirect loop. Although I don't think we need a fallback section since we cache everything.

MoOx:
This should be off by default. Not sure it's a good idea for huge website (with hundreds of pages).
And what are you caching, html pages?

Not experienced with appcache, so can't tell what is better for the FALLBACK. I only use this once https://github.com/putaindecode/tab/blob/gh-pages/cache.appcache

thangngoc89:
Okie. I'll remove it from boilerplate. How about instead of accepting a boolean in setting. We accept a string to pass it to glob. Default to `"**/*.*"`. This way, people can specify their content.

About fallback section, I don't have much experience either.

MoOx:
Good idea for the glob. 
One thing you can do is turning it on for docs :)
And maybe add some documentation somewhere.

thangngoc89:
Yes. Advance usage section . Already has it in mind

thangngoc89:
In the long run. it'll be better to have a service worker by default and fallback to appcache. But writing a proper service worker logic is just too hard.

thangngoc89:
I pushed a new commit.

docs/content/docs/advanced/offline-browsing.md
```diff
@@ -0,0 +1,57 @@
...
+You can config Statinamic to generate a AppCache manifest for you after build.
+
+To enable this feature, add an ``appcache`` field to ``package.json`` under
+``statinamic`` section.
+
+```js
+{
+  ...
+  "statinamic": {
+    "appcache": "**/*.*"
```
MoOx: This will match html and json right?

MoOx:
Docs looks good!

docs/content/docs/advanced/offline-browsing.md
MoOx: Do you think we can find a trick to only get html or json and adjust browsing experience accordingly?

thangngoc89: Yes. It will match everything. deeply nested

MoOx: Maybe that's why FALLBACK is about?

thangngoc89: Only match html and json ? 

`**/*.+(html|json)` . glob uses minimatch under the hood

MoOx: What I am saying is that we got duplicate content here, html and json are the same data. html is supposed to be used for the first render only.

thangngoc89: Ah. FALLBACK is for online required resources.

Example : 

```
FALLBACK
/offline.html /some-file.php
```

Go to /some-file.php when offline, /offline.html will show up instead. I set all the fallback to /index.html to display the nice `404 Not Found Page`

MoOx: We should find a way to only store json and not html. Maybe using fallback, we can serve index.html that will still have the correct url and can get the json?

thangngoc89: Nope. As far as I know, we can't adjust the behaviour.

That's why service worker was invented

thangngoc89: :+1: sound great. I'll do some more research

thangngoc89:
Do you know what is that glob for everything extension excerpt `.html` ?

MoOx:
Use globby instead of glob, this will be easier.

thangngoc89:
:+1: much easier

thangngoc89:
I need a little help. Same question with globby

Try some like 

```
["**/*.*", "**/!*.html"]
["**/*.*", "!**/*.html"]
```

None of them is working

thangngoc89:
Changed to 1 index.html and use it as fallback. Will fix tests later. Gotta go

thangngoc89:
@MoOx fixed tests, should be ready now

docs/content/docs/advanced/offline-browsing.md
```diff
@@ -0,0 +1,65 @@
...
+{
+  ...
+  "statinamic": {
+    "appcache": true
+  }
+  ...
+}
+```
+
+Avaliable values for appcache:
```
MoOx: Available*

docs/content/docs/advanced/offline-browsing.md
```diff
@@ -0,0 +1,65 @@
...
+
+- `true`: Use default globby patterns `[ "**/*.*", "!**/index.html", "index.html" ]`
```
MoOx: should be `[ "**/*.*", "!**/*.html", "index.html" ]` ? (eg: 404.html)

docs/content/docs/advanced/offline-browsing.md
```diff
@@ -0,0 +1,65 @@
...
+- `false` or `null`: Disable appcache manifest generation
+- `array`: Specify your own globby patterns. Be careful, it's all yours.
+- `string`: like array, will be converted to array. Ex: `"foo"` --> `["foo"]`
+
+Statinamic uses [globby](https://www.npmjs.com/package/globby) for matching files in
+``dist`` folder.
+
+Checkout [globby documentation for more information](https://www.npmjs.com/package/globby)
+
+Now rebuild your website and you will noticed a new ``manifest.appcache``
```
MoOx: you will notice*

docs/content/docs/advanced/offline-browsing.md
```diff
@@ -0,0 +1,65 @@
...
+file your ``dist`` folder.
+
+> **NOTE**: AppCache will be disabled in development mode
```
MoOx: `will **not be enabled** during development`

docs/content/docs/advanced/offline-browsing.md
MoOx: Maybe we should explain why we choose this default patterns:

```
This pattern means that only json files will be downloaded. Downloading HTML files too is not useful since appcache will have a proper FALLBACK option to redirect all URLs (when offline) to a single entry point that will start the client app (which will consume json).
```

src/configurator/index.js
```diff
@@ -181,6 +183,31 @@ export default function config(pkg = {}, argv = process.argv) {
...
+  }
+  else if (config.appcache === null) {
+    config.appcache = false
+  }
+  else if (
+    !Array.isArray(config.appcache) &&
+    typeof config.appcache !== "boolean"
+  ) {
+    errors.push(
+      `You provided a(n) ${ typeof config.appcache } for 'appcache' option. ` +
```
MoOx: add single quotes around the variable?

MoOx:
Can you also add one line of doc in Configuration page?

thangngoc89:
Fixed them all. Add a 404.html fixture to cover the default globby.

PR #194, opened 2016-02-25
0.8.1

MoOx:


thangngoc89:
:+1: lgtm

src/builder/server.js
```diff
@@ -125,56 +115,76 @@ export default (webpackConfig, options = {}) => {
...
-
-        if (!firstRun) {
-          cleanNodeCache(config.cwd)
-        }
-        firstRun = false
 
-        urlAsHtml(item.__url, {
-          exports: options.exports,
-          store: options.store,
+      // try 404.html is there is any
```
thangngoc89: * if there is any

MoOx: good catch haha

thangngoc89:
Found that invalid checksum on 404 page. but in my case this happens on both firefox and chrome

MoOx:
Will open an issue for that.

PR #197, opened 2016-02-25
Don't split webpack error and warning to mutiltple lines

thangngoc89:
``debug`` will do it for us.

This is bad right ?

image


MoOx:
Nice.

PR #199, opened 2016-02-25
Use RedBox to display error stack on server

### LINKED ISSUE #196: Dev server render error stack with redbox
Currently, dev server only display the error log, not the full stack. We should throw the full stack on the screen.

Even better if we use [redbox](https://www.npmjs.com/package/redbox-react) to render it. 

>  you can use all the wasted space where your app would be if it didn’t crash to display the error that made it crash

thangngoc89:
screenshot from 2016-02-26 00 32 22

It's a good way to use all the wasted space

Close #196


thangngoc89:
Added tests.

PR #200, opened 2016-02-26
Fix enhance-collection

thangngoc89:


src/enhance-collection/__tests__/index.js
```diff
@@ -32,9 +40,63 @@ test("statinamic/lib/enhance-collection", (t) => {
...
+    [
+      {
+        k: "ey",
+        l: "hi",
+      }
+    ],
+  )
+})
+
+test("multiple filters", (t) => {
```
thangngoc89: @MoOx is this a valid test case for `filters` option ?

MoOx: yeah? What can be the issue?
We should also test that the array can be object or function too.

thangngoc89: The result is wrong.

thangngoc89: When I set filters: `["q", "k"]`, I suppose it will print any object that contains both `q` and `k` key

MoOx: filtering by an array is an && not and ||.

MoOx: result is the only object that contains both right?

thangngoc89: result

```
{ k: 'ey', l: 'hi' }
{ k: 'eye' }
```

thangngoc89: I do suppose it is `&&`

MoOx: Haha ok test is failing I missed that. So yes I guess we have to fix this according to the result you used.

thangngoc89: Yup. I'll fix that. Just ask this here to make sure I understand the code correctly :)

MoOx:
Do you need a hand to fix this?

thangngoc89:
Yes please. No idea where to get started. Feel free to use this branch as start point

MoOx:
On it.

Issue #103, opened 2016-01-26
add in faq "webpack chunk over static assets"


MoOx:
Ref https://github.com/MoOx/statinamic/pull/94/files#r50816037

And at the same time, write doc for static assets folder (and webpack solution)

MoOx:
Closing as it's a pretty shitty option atm. Please track #147 or #356.

PR #204, opened 2016-02-26
Fixed: 404 page overide webpack hmr route

thangngoc89:
Router has catch all route * so webpack hmr route is never reached. 

Why put it above static assets route ? See #103

Ref: https://github.com/glenjamin/webpack-hot-middleware/issues/26#issuecomment-144941322

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+- Fixed: 404 page overide webpack hmr route
+
 # 0.8.1 - 2016-02-25
 
 - Fixed: URLs which does not have the required trailing slash are now adjusted.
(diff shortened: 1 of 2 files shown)


MoOx:
How did you track this?

thangngoc89:
What do you mean by "tracking" ?

thangngoc89:
Reproduce: `npm run test-boilerplate && cd boilerplate && npm start`

MoOx:
I don't have this issue with test-boilerplate or boilerplate (you mixed both in your step).
Not sure what to think :/

thangngoc89:
Actually, I created a new project with `statinamic setup` yesterday (using instruction from docs) and ran into this issue immedialy. So I think the test command may work

PR #212, opened 2016-02-27
Update documentation.

ben-eb:
Dropped the sections on removing linting tools - we shouldn't recommend this.

### DIFF
--- a/docs/content/docs/advanced/engine.md
+++ b/docs/content/docs/advanced/engine.md
@@ -3,15 +3,14 @@ title: Explanation about how the engine works
 incomplete: true
 ---
 
-During the build process, markdown files are parsed, and transformed as JSON
-files (one part with the YAML header meta informations, one part with the body
-as HTML).
-So both client and server can easily consume markdown files (as json) in order
-to render pages from React components. That's it !
+During the build process, markdown files are transformed into JSON; one part
+with the YAML metadata, and one part with the HTML body. So both client and
+server can easily consume these markdown files (as JSON) in order to render
+pages from React components.
 
-On top of that you can add pretty much whatever you want to write styles the way
-you like (CSS, inlines styles...) and you can even create your own pages from
-plain React components. Statinamic is just an helper.
+On top of that, you can easily add your own logic to style your pages the way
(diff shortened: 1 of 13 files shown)


docs/content/docs/advanced/engine.md
```diff
@@ -3,15 +3,14 @@ title: Explanation about how the engine works
 incomplete: true
 ---
 
-During the build process, markdown files are parsed, and transformed as JSON
-files (one part with the YAML header meta informations, one part with the body
-as HTML).
-So both client and server can easily consume markdown files (as json) in order
-to render pages from React components. That's it !
+During the build process, markdown files are transformed into JSON; one part
+with the YAML metadata, and one part with the HTML body. So both client and
```
thangngoc89: @MoOx should we document about using JSON, TOML in front matter ?

user can just using JSON.stringify instead of something like this https://github.com/stylelint/stylelint.io/blob/master/scripts/process-docs-lib/toYAML.js ?

Also hugo (another static site generator) using TOML as default front matter.

docs/content/docs/usage/layouts.md
```diff
@@ -38,7 +38,7 @@ layout: MyCustomLayoutComponent
 _This layout is optional_ and a `PageContainer` contains a minimal fallback if
 it is not available.
 
-[Check out the default PageError and the props it got.](https://github.com/MoOx/statinamic/blob/master/boilerplate/web_modules/layouts/PageError/index.js)
+[Check out the default PageError and the properties it has.](https://github.com/MoOx/statinamic/blob/master/boilerplate/web_modules/layouts/PageError/index.js)
```
thangngoc89: I think this should be `props` . That's a React thing.

ben-eb: :+1:, fixed.

docs/content/docs/advanced/engine.md
MoOx: since we are using https://www.npmjs.com/package/gray-matter I guess we should. Opening a new issue about that.

MoOx:
Thanks a lot @ben-eb!

ben-eb:
You're welcome! :smiley:

PR #217, opened 2016-02-28
Refactor md-collection-loader

### LINKED ISSUE #209: remove the "JSON.stringify" shit in md-collection-loader config
query can be cased as js now?

thangngoc89:
- Pluggable markdown parser (can be ascii parser or something else)
- Move loader query to webpack option (avoid json.stringify)

Close #209

@ben-eb you can use remark now

---
Update:

statinamic requires `mdParser` field : 

```js
(string) => parsedHtml
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+- Changed: Refactor md-collection-loader
+  - Pluggable markdown parser (can be ascii parser or something else)
+  - Move loader query to webpack option (avoid json.stringify)
+
+  See ([#217](https://github.com/MoOx/statinamic/pull/217)) for more information
+
 # 0.8.2 - 2016-02-27
 
 - Fixed: 404 page overide webpack hmr route
--- a/boilerplate/scripts/webpack.config.babel.js
+++ b/boilerplate/scripts/webpack.config.babel.js
@@ -13,32 +13,12 @@ export default {
     loaders: [
       { // statinamic requirement
         test: /\.md$/,
-        loader: "statinamic/lib/md-collection-loader" +
-          `?${ JSON.stringify({
-            context: path.join(config.cwd, config.source),
-            feedsOptions: {
-              title: pkg.name,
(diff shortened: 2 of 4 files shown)


ben-eb:
Looks good, how would I set this option?

thangngoc89:
You have to wait for this PR (there is failing tests right now).

mdParser field requires a function that return a parser HTML content. Use it like this

```js
statinamic: {
  ...
  mdParser: function (fileContent) {
    // do something here
   return parsedHTML
  }
}
```

ben-eb:
:+1:

MoOx:
Some required changes:

- md-collection-loader => text-collection-loader (or just collection-loader)?
- allow from "statinamic.collection" or (textCollection) webpack config, not just "statinamic" (we never know what we will add under statinamic.
- let the possibility to use loader query (ex: I want to use .md and .txt or .html or whatever that is text with a front matter) as it was (so `{ ...loaderUtils.parseQuery(this.query), ...this.options.statinamic && this.options.statinamic.collection }` with validation on the options like for configurator, since this part of statinamic is really important.

We should also doc all this loader in "Configuration" section of the doc.

MoOx:
@thangngoc89 now that you are a collab, please make PR from branch of this repo not your fork, so I can work with you on that :)

thangngoc89:
Should i close this pr and make a new one ?

thangngoc89:
I think this add more complexity for collection-loader. We should instead let user write their own loader for different file type

MoOx:
Why does this add more complexity?
This loader (as you modified it) read text and transform it to html via the cb provided by your config. So not coupled to md at all. Just coupled to the fact that the file must contain a front matter (so is a text file).

For the PR, as you want, I won't work on statinamic this afternoon, maybe this evening (in 6 hours, as I write this).

thangngoc89:
Thanks. That's fair enough. I'll close this PR and send a new 1 from this repo.

PR #279, opened 2016-03-10
Remove redux-thunk

thangngoc89:
Close #256


MoOx:
Are you sure it's not used by the promise middleware for the client side ?

MoOx:
(I have to admit I grab this from a old project where I didn't setup redux + thunk + custom promise middleware)

thangngoc89:
yes. I'm pretty sure. The promise middleware is dispatching action itself https://github.com/MoOx/statinamic/blob/016085d7e74f4783267a7267f6868d9a37c8dc02/src/redux/createStore/index.js#L25-L26 

thunk middleware is just another middleware that dispatch action too.

PR #284, opened 2016-03-10
Fix: url with anchors are being replaced with url without anchors

### LINKED ISSUE #283: url with anchors are being replaced with url without anchors
Eg: go to http://putaindecode.io/en/articles/js/react/unit-testing/, click on a title anchor. Link works, but anchor is quickly stripped from the url bar

thangngoc89:
Close #283

### DIFF
--- a/src/PageContainer/component.js
+++ b/src/PageContainer/component.js
@@ -8,14 +8,14 @@ import urlify from "../_utils/urlify"
 // so we need to normalize according to collection data
 const splatToUrl = (string) => ("/" + urlify(string))
 
-const isDevelopment = process.env.NODE_ENV !== "production"
-const isClient = typeof window !== "undefined"
-const isDevelopmentClient = isDevelopment && isClient
+const isDevelopment = () => process.env.NODE_ENV !== "production"
+const isClient = () => typeof window !== "undefined"
+const isDevelopmentClient = () => isDevelopment() && isClient
 
 let catchLinks
 let browserHistory
 
-if (isClient) {
+if (isClient()) {
   catchLinks = require("../_utils/catch-links").default
   browserHistory = require("../client").browserHistory
 }
@@ -89,7 +89,7 @@ class PageContainer extends Component<DefaultProps, Props, void> {
   }
 
   catchInternalLink() {
-    if (!isClient) {
+    if (!isClient()) {
       return
     }
 
@@ -121,13 +121,12 @@ class PageContainer extends Component<DefaultProps, Props, void> {
     }
 
     const pageUrl = splatToUrl(props.params.splat)
-    if (isDevelopmentClient) {
+    if (isDevelopmentClient()) {
       console.info(`statinamic: PageContainer: '${ pageUrl }' rendering...`)
     }
 
     const item = find(context.collection, pageUrl)
-
-    if (isClient && item) {
+    if (isClient() && item) {
       // adjust url (eg: missing trailing slash)
       const currentExactPageUrl = window.location.href
         .replace(
@@ -139,8 +138,10 @@ class PageContainer extends Component<DefaultProps, Props, void> {
           ),
           "/"
         )
+        .replace(window.location.hash, "")
+
       if (currentExactPageUrl !== item.__url) {
-        console.log(
+        console.info(
           `statinamic: PageContainer: ` +
           `replacing by '${ currentExactPageUrl }' to '${ item.__url }'`
         )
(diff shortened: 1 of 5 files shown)


MoOx:
I would like to cover this with test this time :)

src/PageContainer/component.js
```diff
@@ -100,8 +100,10 @@ export default class PageContainer extends Component {
     }
 
     const item = find(context.collection, pageUrl)
-
+    console.log(isClient)
+    console.log(typeof window !== "undefined")
```
thangngoc89: I injected window to global var

@MoOx any idea why 

```
console.log(isClient) --> false
console.log(typeof window !== "undefined") --> true
```

MoOx: You define window after importing the file, and isClient is defined at the top of the file.

thangngoc89: Got it. So change isClient to a function then ?

anyway, I need to pull babel-plugin-rewire browserHistory. So this PR is blocked by #286

MoOx: Or maybe refactor the code to pass this as props.
And isClient can be replaced by its value (simpler to test)

thangngoc89: Yeah, that's easier. Are you ok with pulling babel-plugin-rewire ? Do you have any refer method ?

MoOx: Looks like a bad design to need this thing.
What about finding a way to pass this as a prop?

MoOx: (Or via the context)

thangngoc89: We only use browserHistory in PageContainer so no need for context. So browserHistory, isClient, ... will be passed via props

MoOx: passed via props means that something is gonna pass those thing.
isClient is not really useful, just drop it an use a simple test directly in the component.

thangngoc89: We have react-redux connect to pass those via props.

So using `typeof window !== 'undefined'` directly right ?

MoOx: Redux will be removed one day or another :)

thangngoc89: Ouch. browserHistory required may DOM API, can't mock them all

MoOx: We will need jsdom one day or another. We should add it now.

MoOx: First, double check that jsdom support browser history :/

thangngoc89: On it

thangngoc89:
Ouch. jsdom doesn't work with `history` package. `history` requires `window.sessionStorage` which jsdom [doesn't have](https://github.com/tmpvar/jsdom/issues/1137)

MoOx:
What about mocking window requirements?!

thangngoc89:
I'm mocking it. But history package uses a large browser API. window.sessionStorage wasn't the only thing it required.

```
SecurityError
    HistoryImpl._sharedPushAndReplaceState (/home/khoa/web/statinamic/node_modules/jsdom/lib/jsdom/living/window/History-impl.js:87:15)
    HistoryImpl.replaceState (/home/khoa/web/statinamic/node_modules/jsdom/lib/jsdom/living/window/History-impl.js:72:10)
    History.replaceState (/home/khoa/web/statinamic/node_modules/jsdom/lib/jsdom/living/generated/History.js:89:58)
```

MoOx:
So maybe mocking browser-history like your proposed before...?

thangngoc89:
I managed to make it work with jsdom :+1: . Setup is complete. Now it is the real testing part

MoOx:
This is reallyyyy nice. Can you just add the interesting test ? :D

thangngoc89:
@MoOx  done

src/PageContainer/component.js
```diff
@@ -6,14 +6,14 @@ import urlify from "../_utils/urlify"
...
+const isClient = () => typeof window !== "undefined"
+const isDevelopmentClient = () => isDevelopment() && isClient
 
 let catchLinks
 let browserHistory
 
-if (isClient) {
+if (isClient()) {
   catchLinks = require("../_utils/catch-links").default
   browserHistory = require("../client").browserHistory
```
thangngoc89: @MoOx here is why I need to clean node cache. I need to refresh `browserHistory`

MoOx: oh, ok.

MoOx:
I rebase this ;)

thangngoc89:
Thanks

PR #291, opened 2016-03-11
Add flowtypes

### LINKED ISSUE #282: Add Flow for better code quality
http://flowtype.org/

thangngoc89:
Close #282 

Known limitation of flow

- Can't use arrow function (not flowtypes fault. babylon doesn't understand arrow function with type annotate)


.travis.yml
```diff
@@ -15,7 +15,9 @@ cache:
     - docs/node_modules
 before_script: npm prune
 
-script: npm run test-with-coverage
+script:
+  - npm run flow:check
+  - npm run test-with-coverage
```
MoOx: can you instead add flow:check in test? maybe postlint ?

thangngoc89: Flowtype only supports *nix

MoOx: Wow. So I will prefer a script that run flow type depending on the platform so when I run "npm test", I don't miss flow check.

thangngoc89: Ok. That should be easy. But I will wait for fix/windows branch to be merged before implementing this to avoid conflict (most of code was written overthere)

MoOx: Sure!

thangngoc89:
A little bit stuck. Not sure how to fix this 

https://travis-ci.org/MoOx/statinamic/builds/115493140#L250-L254

```
src/static/to-html/index.js:106
106:         .then((html) => writeHTMLFile(filename, html))
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ function call. Callable signature not found in
106:         .then((html) => writeHTMLFile(filename, html))
                             ^^^^^^^^^^^^^ Promise
```

Ref: https://github.com/facebook/flow/issues/1408

MoOx:
Not sure how to fix that neither :(

thangngoc89:
In the bright side. I rewrote the writeHTMLFile function to use promise instead of sync operation

thangngoc89:
Stupid flowtype:

```
// define it
config.appcache: Array<string> | false

// Function
writeAppcache(pattern: Array<string>)

// Now call it

if (
    config.appcache &&
    typeof config.appcache !== "boolean"
  ) {
    writeAppcache(config.appcache)
}
```

And it's throw error 

```
boolean literal `false`. This type is incompatible with array type
```

thangngoc89:
Eslint is complaining about component's properties order

https://travis-ci.org/MoOx/statinamic/builds/115573573#L365-L365

Flowtypes should be declared at top of class. What do you think ? Remove that rule ?

thangngoc89:
I added flowtype to any file that I can add. There is some gotchas but most of them is fixable. But there is some that required a large refactor of our code base (see above comment for info).

This PR is ready for review @MoOx

src/static/postbuild.js
```diff
@@ -3,7 +3,7 @@ import color from "chalk"
 import { writeFile } from "fs-promise"
 import writeAppcache from "../_utils/appcache"
 
-export default (config, files, log) => {
+export default function(config, files, log) {
```
MoOx: if seems that flow can handle arrow function already, what is the error that we got here?

thangngoc89: It is babel who doesn't understand arrow functiom with flowtypes

MoOx: Is there an open issue about that?

thangngoc89: https://phabricator.babeljs.io/T2400

https://phabricator.babeljs.io/T2612

PR #300, opened 2016-03-14
0.9.0

MoOx:
0.9 is almost ready, we still miss

- [x] #276 - @thangngoc89
- [x] #262 - @MoOx
- [x] #269 - @MoOx
- [x] #279  - @thangngoc89
- [x] #325 - @thangngoc89
- [x] #324 - @thangngoc89
- [x] check docs
- [x] tldr section in CHANGELOG
- [x] allow install from github #203

PR #314, opened 2016-03-15
Backport master changes to next

MoOx:


MoOx:
Well maybe not the best idea I had.

thangngoc89:
This is easier than the other way.

MoOx:
The other way is #300 and it's not easier since they are conflicts, the same conflicts.

MoOx:
Not sure if merging master into next and then merging next into master make more sense.
I will unprotect and rebase next on master.

thangngoc89:
I won't mess with #300. That's a can of worms

PR #316, opened 2016-03-15
Correct nyc config

thangngoc89:


MoOx:
Seriously? wtf. Anyway thanks!

PR #322, opened 2016-03-17
Fix: open static server after build

thangngoc89:


MoOx:
If you say so, I don't even use this feature.

MoOx:
That said, after 0.9, I won't accept any commit without a test to prove the thing ^^

thangngoc89:
> That said, after 0.9, I won't accept any commit without a test to prove the thing ^^

flow would help us here. but `src/bin/server.js` needs a lot of love to be able to use flow.

PR #324, opened 2016-03-18
Update dependencies

thangngoc89:
- Move simple-json-fetch to dep (missing when install from boilerplate)
- Allow peerDep whatwg-fetch >= 0.9.0 (React is using 0.11.0, no reason to use older one)
- Remove react-jsx-.... (not needed anymore)
- Added missing babel-eslint to opPeerDeps
- Move remark* from devDep to opPeerDeps
- Remove hightlight.js from opPeerDeps (became a dep of remark-highlight.js)

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "remark": "^4.1.1",
     "rimraf": "^2.4.2",
     "rss": "^1.2.0",
+    "simple-json-fetch": "^1.0.1",
     "strip-markdown": "^0.3.1",
     "underscore.string": "^3.2.3",
     "valid-url": "^1.0.9",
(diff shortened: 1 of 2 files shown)


thangngoc89:
Well. There is some missing deps: remark and remark-*

MoOx:
The problem is: from our project it's (from what I know) impossible to test deps that should be in deps, because to run test we use devdeps so if a dep is in devdeps instead of deps, we just don't know it.
Anyway, we should be careful and try ourself  from git before cutting a release.

thangngoc89:
> from our project it's (from what I know) impossible to test deps that should be in deps

I am aware of that. I'm  starting a fresh new project using latest commit in next branch.

I'll update this PR with all missing deps that I can find

thangngoc89:
I updated the PR. Let's hope I didn't break anything.

package.json
```diff
@@ -125,9 +124,10 @@
     "react-router": "^2.0.0",
     "redux": "^3.0.0",
     "webpack": "^1.12.1",
-    "whatwg-fetch": "^0.9.0"
+    "whatwg-fetch": ">=0.9.0"
```
MoOx: Can you just update to 0.11.0 ?

thangngoc89:
PR updated

PR #326, opened 2016-03-19
Fix content loader

### LINKED ISSUE #325: [next] context-loader bugs
There is 2 bugs with latest content-loader: 

- Default loader should not be required [at the top of the file](https://github.com/MoOx/statinamic/blob/81baba84c5b8a49564ef8b2c82996148bf8f02ed/src/content-loader/index.js#L12). This would require user install remark-* while they don't need them. 

- webpack's loader parser remove all function from query ==> no custom engine

I suggest we ditch webpack loader's query completely and use a custom field in webpack config

thangngoc89:
close #325

### DIFF
--- a/src/content-loader/index.js
+++ b/src/content-loader/index.js
@@ -9,7 +9,6 @@ import feed from "./feed"
 import cache from "./cache"
 import description from "./description"
 import validator from "./validator"
-import defaultRenderer from "./default-renderer"
 
 let timeout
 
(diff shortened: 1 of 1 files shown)


src/content-loader/index.js
```diff
@@ -9,12 +9,14 @@ import feed from "./feed"
...
-import defaultRenderer from "./default-renderer"
 
 let timeout
 
 module.exports = function(input) {
-  const query = loaderUtils.parseQuery(this.query)
+  const query =
+    this.options.statinamic
+      ? this.options.statinamic.loader
+      : loaderUtils.parseQuery(this.query)
```
MoOx: I will add a merge instead, as well as test.

thangngoc89: OK. 

```js
{
  ...this.options.statinamic.loader,
  ...loaderUtils.parseQuery(this.query)
}
```

right ?

I'll do it tmr.

MoOx: yeah that should be good :)

PR #331, opened 2016-03-21
Ensure .gitignore when setup a new boilerplate

### LINKED ISSUE #323: .gitignore is renamed to .npmignore in boilerplate
This issue from npm was there since 2011 (https://github.com/npm/npm/issues/1862)

For those who wanna contribute to statinamic. This is a easy patch. Just rename .npmignore to .gitignore after boilerplate is copied

Ref: https://github.com/MoOx/statinamic/blob/81baba84c5b8a49564ef8b2c82996148bf8f02ed/src/bin/commands/setup.js#L51-L52

thangngoc89:
Close #323

### DIFF
--- a/npm/postinstall.js
+++ b/npm/postinstall.js
@@ -2,11 +2,30 @@ const stat = require("fs").stat
 const spawn = require("child_process").spawn
 const join = require("path").join
 const pkg = require("../package.json")
-
+const fs = require("fs-extra")
 // no need for th is step on CI
 if (process.env.CI) {
   process.exit(0)
 }
+
+const boilerplateDir = join(__dirname, "../boilerplate")
+
+stat(join(boilerplateDir, ".npmignore"), function(err) {
+  if (err) {
+    console.log("No .npmignore in boilerplate to rename")
+    return true
+  }
+
+  fs.move(
+    join(boilerplateDir, ".npmignore"),
+    join(boilerplateDir, ".gitignore"),
+    function(err) {
+      if (err) {
+        throw new Error("Cannot rename .npmignore to .gitignore in boilerplate")
+      }
+    }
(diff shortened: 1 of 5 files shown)


npm/postinstall.js
```diff
@@ -2,12 +2,30 @@ const stat = require("fs").stat
 const spawn = require("child_process").spawn
 const join = require("path").join
 const pkg = require("../package.json")
-
+const fs = require("fs-extra")
```
MoOx: can't you use fs-promise here too?

npm/postinstall.js
```diff
@@ -2,12 +2,30 @@ const stat = require("fs").stat
...
 // no need for th is step on CI
 if (process.env.CI) {
   process.exit(0)
 }
 
+const boilerplateDir = join(__dirname, "../boilerplate")
+
+stat(join(boilerplateDir, ".npmignore"), function(err, stats) {
+  if (err || stats.isDirectory()) {
+    return true
```
MoOx: why do you return true here?

MoOx: maybe just if (err) console.log("no npmignore in boilerplate to rename)?

MoOx: no need to check directory

thangngoc89: It's just for escape the function. coding style.

npm/postinstall.js
thangngoc89: Why ? fs-promise is just a wrapper of fs-extra. I'm using async operation here :). And I don't like a mix of callback and promise.

thangngoc89:
PR updated

MoOx:
Going to add commit + release as 0.9.1

Issue #301, opened 2016-03-14
Performance improvement


MoOx:
We really need to discuss that. Since we started using BABEL_DISABLE_CACHE, everything is super slow.

First, using BABEL_DISABLE_CACHE for the static build is brutal: we should instead just "remove" the cache before doing the build, since during the build we are not modifying the files and can benefit from the cache, this should speed up things a lot (I hope).

Second, during development, maybe we should consider trying (again) a more smart cache cleaner that will remove (eg: when a css file change) the appropriate cache _only_ (not everything).
Maybe we should handle that in babel-plugin-webpack-loaders (poke @istarkov).

Any thoughts?

istarkov:
As a fast idea to use something like this (for css, sass) instead of `BABEL_DISABLE_CACHE`

```bash
egrep -irl --include=*.{js,jsx} "(.css\'|.sass\')" $BABEL_CACHE_PATH | xargs rm
```

_where $BABEL_CACHE_PATH must be set to cache directory_

istarkov:
But I'm not even sure that babel uses different files for cache ;-)

istarkov:
Looks like it's just a single json file https://babeljs.io/docs/usage/require/

istarkov:
May be this helps
```
# touch files to change access times, so if cache uses this information this can help
egrep -irl --include=*.{js,jsx} "(.css\'|.sass\')" ./src | touch
```

MoOx:
Workaround: use a local babel cache directory, and remove it before static build (but the issue for the live server will still occurs, we might delete the babel cache on each prerender, that might make things still slow, but maybe a little less slower than it is for now)

Possible fixes:
- reconsider using a webpack builded file for the server, but this time with nodemon to reload the server when a file changes. That will make us have 2 servers: one for webpack build (client/server), one that use webpack server build refreshed via forever/nodemon. Now that there is less boilerplate, we might reconsider this option.
- drop the prerendering or make it optional. Maybe making this optional will make things faster with not a lot of effort.

thangngoc89:
The first option is recommended (or was recommended) by webpack https://github.com/webpack/react-webpack-server-side-example

thangngoc89:
This might be useful: https://github.com/trivago/parallel-webpack

We can use this approach to have static and client build in parallel.

MoOx:
This is now my top priority and here is what I think we should do:

- remove dev server rendering for now
- have a fast start for dev server
- have a fast static build

[...]

PR #336, opened 2016-03-22
Use babel-register instead of babel-core/register

### LINKED ISSUE #335: Missing babel-register in dep
We only have babel-core, not babel-register so this will break stuff.

This is also a good chance to replace `babel-core/register` with `babel-register`. (will be deprecated soon)

thangngoc89:
Close #335

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+- Fixed: Missing babel-register package.
+  ([#335](https://github.com/MoOx/statinamic/issues/335))
+
 # 0.9.1 - 2016-03-21
 
 - Fixed: default boilerplate have a correct .gitignore file (not .npmignore)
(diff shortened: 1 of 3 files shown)


thangngoc89:
This must be release at 0.10.0 (breaking change)

MoOx:
Why this should be  a breaking change? It's used internally only. Should just be a dep.

thangngoc89:
> Why this should be a breaking change?

I added a peer dep.

> It's used internally only. Should just be a dep.

we don't put babel stuff in dep. I'm trying to avoid duplicate package here.

MoOx:
Stuff used internally should be in deps. Anyway I don't think we are going to use this  for long (ref #301).
Please put in dep so we can release as a fix.

thangngoc89:
PR updated.

PR #338, opened 2016-03-22
docs: Fix shell command highlight

thangngoc89:
3 aliases for this: bash, sh, zsh.

### DIFF
--- a/docs/content/contributing/index.md
+++ b/docs/content/contributing/index.md
@@ -6,7 +6,7 @@ title: "How to contribute to Statinamic"
 
 To work on the boilerplate, run the following:
 
-```console
+```sh
 cd boilerplate
 ln -sf ../docs/package.json
 ln -sf ../docs/node_modules node_modules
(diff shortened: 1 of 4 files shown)


MoOx:
This aliases are for shell scripts. This snippets are not scripts. These are terminal (console) commands so it's the correct aliases from what I know.

thangngoc89:
This is no styling in docs right now 

http://moox.io/statinamic/docs/setup/

See https://github.com/isagalaev/highlight.js/blob/master/src/languages/bash.js

MoOx:
It's an issue with the highlighter, not the syntax itself. For example on github it's fine https://github.com/MoOx/statinamic/blob/add524b5ae0f3907080c6049498fdb34c61d4ba1/docs/content/docs/setup.md#create-a-new-folder (even if there is not a lot of colors.

thangngoc89:
Github does support the console keyword https://github.com/github/linguist/blob/d9914307eb4dc306d8312d179cb75c78202f7033/lib/linguist/languages.yml#L3360-L3368

> even if there is not a lot of colors

That's extracly how it will be after this PR. I'm fine with closing this PR (not much colors)

MoOx:
Maybe we should open one on highlight.js ? FYI, console and shell does not produce the same render from what I know.

thangngoc89:
Github calls it shell section.

> FYI, console and shell does not produce the same render from what I know.

Yeah.

MoOx:
Closing in favor of https://github.com/isagalaev/highlight.js/issues/1126

PR #339, opened 2016-03-22
Faster npm install on CI without progress

MoOx:
Should be faster


thangngoc89:
Yeah. I disabled this locally a long time ago.

PR #343, opened 2016-03-23
Add service worker

### LINKED ISSUE #153: Smart Offline mode via Service Worker.

thangngoc89:
# Review needed

This is my first time with Service worker. Please review the SW logic carefully :).

The PR contains many unrelated files to SW, if you aren't familiar with Phenomic codebase, please take time to review these files 

https://github.com/MoOx/phenomic/blob/feat/sw/src/_utils/offline/service-worker/sw.template.js
https://github.com/MoOx/phenomic/blob/feat/sw/src/_utils/offline/service-worker/sw-register.template.js

By the way, the current behavior of Service worker doesn't different from AppCache. If you have a better sw logic for a static site, please give me a shout

---
TODO:

- [x] Add tests
- [x] Merge appcache and service worker into offline section (they share the same pattern, setting)
- [x] Document

---

Changes:

- Initial work for SW. Right now, the default SW template isn't different than the "stupid" appcache options but in the future, I want user to be able to provide their own SW template 
- Add a deprecated warning when use `appcache` option. **BREAKING CHANGE:** removed alibity to define custom globby pattern. Use offline option to define it

- Add `offline` option. It accepts: 

  - `boolean`: default = false (enable/disable offline features)
  -  `object` contains these keys:
    - `appcache: boolean = true` (enable/disable appcache)
    - `serviceWorker: boolean = true` (enable/disable serviceWorker)
    - `pattern: string | array = [ "**", "!**/*.html", "index.html" ]` : globby pattern

Close #153


thangngoc89:
@MoOx I updated #1 post. Could you please give some feedback (ignore the messy code for now)

src/configurator/validators/offline.js
```diff
@@ -0,0 +1,61 @@
...
+  pattern: defaultPattern,
+}
+
+export default (
+  { config, errors }:
+  { config: StatinamicConfig, errors: Array<string>}
+) => {
+  // Disable offline for development
+  if (config.dev) {
+    config.offline = false
```
MoOx: add a warning/info here so people knows :)

src/configurator/validators/offline.js
```diff
@@ -0,0 +1,61 @@
...
+    return
+  }
+
+  if (!config.offline) {
+    return
+  }
+
+  if (!config.offlineConfig) {
```
MoOx: why not just offline: Object | true ? true => default config, Object means true.

MoOx:
Could you add fallback for previous appcache option + add a warning so people can update without having to handle the change of config? I will push some code like this for layouts definition changes that will still support old method for a moment.

src/configurator/validators/offline.js
thangngoc89: Do you remember our chat on gitter about `flow` doesn't allow type mutation ? 

for ex:

```js
// define it
config.appcache: Array<string> | false

// Function
writeAppcache(pattern: Array<string>)

// Now call it

if (
    config.appcache &&
    typeof config.appcache !== "boolean"
  ) {
    writeAppcache(config.appcache)
}
```

thangngoc89:
> add fallback for previous appcache option

good idea

src/configurator/validators/offline.js
MoOx: Yeah but we just cannot provide a weird UX because of flow. We have to find a workaround.

thangngoc89: OK. how about the configurator will convert 

```
offline: Object
```

to 

```
offline: true
offlineConfig: Object
```

This will sastify flow and also a clean logic for the core

MoOx: why not.

MoOx: hum maybe we can just consider test (typeof offline ==="object") instead of doing (if offline) ?

MoOx: I don't want to add weird technical debt because of flow being annoying on a common pattern for conf (a value means it's enabled)

thangngoc89: > (typeof offline ==="object") 

Not sure what do you mean here. Will take a look at config section tmr. I wrote those LoC 2 weeks ago.

MoOx:
Would be nice to have this for next version. Do you need help to handle warning for "appcache" option?

thangngoc89:
> Would be nice to have this for next version

:+1: 

> Do you need help to handle warning for "appcache" option?

I can handle this. This PR a quite big and contains some unrelated changes. I'm removing them from this PR and send them in different PR

MoOx:
You can rebase :)

thangngoc89:
> You can rebase :)

With squash and merge button. I can freely click Update branch button right now.

MoOx:
Yeah true :)

thangngoc89:
I did a rebased. Let's hope that I didn't messed up anything

thangngoc89:
@MoOx PR updated. Please take a look

docs/content/docs/advanced/offline-browsing.md
```diff
@@ -16,55 +16,61 @@ on the run without a internet connection.
...
+- [Learn about AppCache](http://www.html5rocks.com/en/tutorials/appcache/beginner/)
 
-### [AppCache browser support from caniuse](http://caniuse.com/#search=appcache)
+- [AppCache browser support from caniuse](http://caniuse.com/#search=appcache)
 
-You can config Phenomic to generate a AppCache manifest for you after build.
+## Service Worker
 
-To enable this feature, add an ``appcache`` field to ``package.json`` under
+This is a part of ES2015 specifications , only modern browsers support it.
```
MoOx: We should mention the fallback provided? (Is there any?)

docs/content/docs/advanced/offline-browsing.md
```diff
@@ -16,55 +16,61 @@ on the run without a internet connection.
...
+
+- [Learn about Service Worker](http://www.html5rocks.com/en/tutorials/service-worker/introduction/)
+
+- [Service worker browser support from caniuse](http://caniuse.com/#search=service-worker)
+
+## How to use
```
MoOx: ## How to enable offline support

docs/content/docs/advanced/offline-browsing.md
```diff
@@ -16,55 +16,61 @@ on the run without a internet connection.
...
 
-Available values for appcache:
+Available values for `offline`:
 
-- `false` or `null`: Default value. Disable appcache manifest generation
-- `true`: Use default globby patterns `[ "**/*.*", "!**/*.html", "index.html" ]`
+- `false`: Default value. Disable `offline` support.
+- `true`: Use default globby patterns `[ "**", "!**/*.html", "index.html" ]`
 
   This pattern means that only json files will be downloaded.
```
MoOx: >    This pattern means that only json files will be downloaded. 

This is not really true, isn't it? What about assets etc? CSS?

docs/content/docs/advanced/offline-browsing.md
```diff
@@ -16,55 +16,61 @@ on the run without a internet connection.
...
-``dist`` folder.
+  Phenomic uses [globby](https://www.npmjs.com/package/globby) for matching files in
+  ``dist`` folder.
 
-Checkout [globby documentation for more information](https://www.npmjs.com/package/globby)
+  Checkout [globby documentation for more information](https://www.npmjs.com/package/globby)
 
-Now rebuild your website and you will notice a new ``manifest.appcache``
-file your ``dist`` folder.
+- `object`: Contains 3 keys. Please aware that you don't need to define all of 3 keys. They will be defined with their default values.
```
MoOx: Please *be* aware.

docs/content/docs/advanced/offline-browsing.md
```diff
@@ -16,55 +16,61 @@ on the run without a internet connection.
...
 
-> **NOTE**: AppCache will **not be enabled** in development mode
+  - `appcache: boolean = true`: Enable/Disable AppCache seperately
+  - `serviceWorker: boolean = true`: Enable/Disable Service Worker seperately
+  - `pattern: Array = [ "**", "!**/*.html", "index.html" ]`: Define your own globby pattern. A useful case for this option is when you need to remove images from caching list if your website is quite big. But be careful, double check everything before you apply the change.
 
-Congratulation. Your website is now a offline-first application.
+Now rebuild your website and you will notice some new files such as  ``manifest.appcache``, ``sw.js``, ``sw-register.js``
```
MoOx: you will notice some > you may notice some

and add "depending on the options you provided" at the end, after sw-register.js

MoOx:
@thangngoc89 I rebased this branch (so force pushed), be sure to pull before fixing things :)

src/_utils/offline/service-worker/sw.template.js
```diff
@@ -0,0 +1,57 @@
+const cacheName = "<%= cacheName %>"
+
+/* eslint-disable quotes */
+const cacheList = JSON.parse('<%= files %>')
```
MoOx: does this list will contains all files? Like those already defined in `window.__COLLECTION__`?

src/_utils/offline/service-worker/sw.template.js
```diff
@@ -0,0 +1,57 @@
...
+
+self.addEventListener("install", (event) => {
+  console.log("[SW]:", "installing")
+  event.waitUntil(
+    caches
+      .open(cacheName)
+      .then((cache) => (
+        cache
+          .addAll(cacheList)
+          // Go from install to active without waiting step
```
MoOx: what step are possible?

src/_utils/offline/service-worker/sw.template.js
```diff
@@ -0,0 +1,57 @@
+const cacheName = "<%= cacheName %>"
+
+/* eslint-disable quotes */
+const cacheList = JSON.parse('<%= files %>')
+/* eslint-enable quotes */
+
+self.addEventListener("install", (event) => {
+  console.log("[SW]:", "installing")
+  event.waitUntil(
+    caches
```
MoOx: where is defined `caches`? It's built-in?

src/builder/index.js
```diff
@@ -19,42 +20,21 @@ export default function(options: Object): void {
...
-        .sort((a, b) => (a.toLowerCase() > b.toLowerCase()) ? 1 : -1)
-        .forEach((name) => {
-          if (name.endsWith(".js")) {
-            assetsFiles.js.push(name)
-          }
-          else if (name.endsWith(".css")) {
-            assetsFiles.css.push(name)
-          }
-        })
+      const assetsFiles = sortAssets(stats.toJson().assetsByChunkName)
```
MoOx: nice idea, in the futur we might make this method customizable

src/configurator/__tests__/assets.js
```diff
@@ -0,0 +1,61 @@
...
+  t.deepEqual(
+    config.assets,
+    {
+      path: join(process.cwd(), config.source, "assets"),
+      route:"assets",
+    }
+  )
+})
+
+test("should not accept false for 'asset' option", (t) => {
```
MoOx: why "not"?

src/configurator/__tests__/assets.js
```diff
@@ -0,0 +1,61 @@
...
+  const config = configurator({ phenomic: { "assets": false } }, [])
+  t.is(config.assets, false)
+})
+
+test("should not accept null for 'asset' option", (t) => {
```
MoOx: same here?

src/configurator/__tests__/assets.js
```diff
@@ -0,0 +1,61 @@
...
+  const config = configurator({ phenomic: { "assets": null } }, [])
+  t.is(config.assets, null)
+})
+
+test("should accept object for 'asset' option", (t) => {
```
MoOx: and here should be "should not"?

src/configurator/__tests__/offline.js
```diff
@@ -0,0 +1,146 @@
...
+    false
+  )
+})
+
+test("should provide default offlineConfig when 'offline' = true", (t) => {
+  t.deepEqual(
+    configurator({ phenomic: { offline: true } }, []).offlineConfig,
+    {
+      serviceWorker: true,
+      appcache: true,
```
MoOx: how browser will behave with appcache + service worker enabled?! Is this not overkill?! Does service worker take precedence?

src/configurator/validators/offline.js
```diff
@@ -0,0 +1,93 @@
...
+      errors.push(
+        "phenomic.appcache option was replaced by phenomic.offline option. " +
+        " You can't define both of them at the same time."
+      )
+    }
+    config.offline = true
+
+    log(yellow(
+      "DEPRECATED: phenomic.appcache option was deprecated " +
+      "and will be removed soon. We updated your configurator to use " +
```
MoOx: > We updated your configurator to use

This message is not good. User might think we updated their package.json.
We should say "We assumed you want to enable offline support, so we used"

src/configurator/validators/offline.js
```diff
@@ -0,0 +1,93 @@
...
+      errors.push(
+        `You provided an '${ typeof config.offlineConfig.pattern }' ` +
+        "for 'phenomic.offline.pattern' option. " +
+        "This option accepts a string or an array."
+      )
+    }
+  }
+  else {
+    errors.push(
+      `Your prodided an '${ typeof config.offline }'` +
```
MoOx: provided

MoOx:
I made a review. This looks good.

Some questions:

- how does browsers behave with appcache and sw enabled?
- Can't we put all js in the bundle? Or at least just add all code for sw in a single script? I understand there is currently 2 files just for sw.
- Can't we reuse files from `window.__COLLECTION__`?

MoOx:
According to MDN

> Furthermore, if you use AppCache and SW on a page, browsers that don’t support SW but do support AppCache will use that, and browsers that support both will ignore the AppCache and let SW take over.

We should mention that in the doc. I found this quote in https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API/Using_Service_Workers#Registering_your_worker (2nd note).

docs/content/docs/advanced/offline-browsing.md
thangngoc89: Yes. Both appcache and service worker have fallback option

src/_utils/offline/service-worker/sw.template.js
thangngoc89: Yes. This list contains all files received from globby

src/_utils/offline/service-worker/sw.template.js
thangngoc89: Yup. It's browser's global

thangngoc89:
> Can't we put all js in the bundle? Or at least just add all code for sw in a single script? I understand there is currently 2 files just for sw.

It's possible to inline sw-register.js file since it's small. We can bundle it into the js bundle too but this involves changing the boilerplate and a webpack plugin.

> Can't we reuse files from `window.__COLLECTION__`

I'm not sure. `sw.js` must be a seperate file. And from what I read, it can't access DOM

thangngoc89:
> how does browsers behave with appcache and sw enabled?

This is the actual behaviors : 

- First view, appcache will be activated. Download all files in appcache.manifest. Service worker will be registered in this phase.

- Second view (reload), appcache will be disabed, service worker will take over.

src/configurator/__tests__/offline.js
thangngoc89: I added a comment about this below.

src/_utils/offline/service-worker/sw.template.js
thangngoc89: A service worker lifecycle: 

image

Waiting step is when there is already a SW in the page and a new SW was registered. The new one have to go to the waiting step until reload.

thangngoc89: I found a better SW lifecycle 

image

src/_utils/offline/service-worker/sw.template.js
thangngoc89: Actually, we don't need JSON.parse here. It can just be like this 

`const cacheList = <%= files %>`

But this will break babel and eslint so I choose this approach

docs/content/docs/advanced/offline-browsing.md
thangngoc89: Mentioned in the default pattern section

thangngoc89:
I updated the PR. Remove almost all fixtures and replaced with mockfs

src/_utils/offline/service-worker/sw.template.js
MoOx: Maybe we could delete the template part and use webpack defineplugin instead?!

MoOx:
> It's possible to inline sw-register.js file since it's small. We can bundle it into the js bundle too but this involves changing the boilerplate and a webpack plugin.

I think one think we can do is inject stuff in webpack config. We could inject sw-register before running webpack in phenomic-client entry.

src/_utils/offline/service-worker/sw.template.js
MoOx: From what I understand, for now, it's not different than appcache: all files (or only files from globby) are cached initially.
So service worker does not bring any value compared to appcache.

One thing we can do (I think it should be easy), is to accept 2 patterns: files that must be cached asap (current behavior) + files that can be cached when viewed.

Eg: putaindecode.io is a (big) website, and I don't want to cache all the files, all posts. But caching only viewed posts is a pretty nice feature, what do you think?

thangngoc89:
OK. Do you know how to inject it into webpack ?

MoOx:
In config we have `config.webpackConfig` and `config.webpackConfigClient`. We can inject the file we want in `config.webpackConfigClient.entry.phenomic-client` (btw, we may validate that this entry name exist, or at least that there is one).

For the injection we can do: 
- if there is a single entry, just append (entry can be a path OR a array of path)
- if there is multiple entries, and one is called phenomic-client, append to this one.
- if there is multiple entries, and no one is called phenomic-client, throw an error for now (if people ask we might allow the name to be changed, but not for now - or if you want, add a param in config like "entryToInjectScripts", default to phenomic-client, cause maybe we might add more for other features).

thangngoc89:
> entry can be a path OR a array of path

Didn't know about this. 

---

I think it's easier to inject to the first entry. sw-register can be load async so it doesn't matter which entry to inject it. as long as it's in the bundle

MoOx:
I don't want people to come and say "you inject something that I don't control". User need to be able to control.

thangngoc89:
> User need to be able to control.

So default to the first entry in the list and provide `entryToInjectScripts` to let people define the entry. Good ?

MoOx:
Will should do what [I proposed before](https://github.com/MoOx/phenomic/pull/343#issuecomment-210287723) or keep it the way it is for now.

thangngoc89:
OK. I'm writing CHANGELOG and merge this PR for now. Will open an issue and revisit this latter.

PR #351, opened 2016-03-25
Fix: Remove unused define env vars in webpack config

thangngoc89:
This should be released as a hotfix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## Boilerplate
+
+- Removed: Remove unused define environment variables in webpack client config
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+- Fixed: Remove unused define environment variables in webpack's config
```
MoOx: It's not really a fix, it's a boilerplate update, can you prefix with "## Boilerplate" and change the message to "- Removed: useless injected env variable to client build" ?

thangngoc89:
PR updated

Issue #357, opened 2016-03-31
boilerplate: PageError layout is complaining about propTypes


thangngoc89:
See docs PageError layout for example implementation

```
Warning: Failed propType: Required prop `error` was not specified in `PageError`. Check the render method of `PageContainer`.
Warning: Failed propType: Required prop `errorText` was not specified in `PageError`. Check the render method of `PageContainer`.
```

MoOx:
How do you access the page error? real 404? dev 404? 404.html?

thangngoc89:
I don't access it. It shows up in the terminal when I run `statinamic start`

MoOx:
from docs? boilerplate? your project?

thangngoc89:
I said shit. Nevermind. You can install a fresh boilerplate. Go to `/404.html` route and you'll see the error.

thangngoc89:
This is the extract commit to fix this https://github.com/thangngoc89/dnh-blog/commit/0cea272206a68bde86f2afb85a233033e327ae15

MoOx:
It's not logical to provide an 404 as a default value for an error page. I did that in the past but reverted (can't remember why). Will try an alternative solution.

PR #372, opened 2016-04-10
Refactoring for layouts + loading indicator

### LINKED ISSUE #182: Add a default loading indicator between page in default boilerplate and docs
Something like [topbar](http://buunguyen.github.io/topbar/)?
A simple spinner?
A content placeholder?
topbar + a content placeholder?

MoOx:
Sorry for the big PR, maybe try to review one commit at a time :)

The refactoring supports current method, so no breaking changes. Show a notice to users.

Loading page looks like this:

screen shot 2016-04-11 at 00 30 27

The spinner in the middle only show if the delay is > 250ms, which avoid flash spinner (which sucks).
Note the top bar (logarithmic effect: start quick, and the more it loads, the less its fast). Better UX imo compared to most loading topbar which quickly load around 20% then go right to the 100%.

Boilerplate 404 is now like docs

screen shot 2016-04-11 at 00 30 24

I will add more components and layouts in default boilerplate after that (have local wip, but I found a bug and want to write test for it).

Will closes #182 and #357.


CHANGELOG.md
```diff
@@ -1,3 +1,47 @@
...
+  import StatinamicPageContainer from "statinamic/lib/PageContainer"
+
+  import Page from "../layouts/Page"
+  import PageError from "../layouts/PageError"
+  import PageLoading from "../layouts/PageLoading"
+  import Homepage from "../layouts/Homepage"
+
+  class PageContainer extends Component {
+    render() {
+      const { props } = this
```
thangngoc89: This is super confusing. Why not just `{ ...this.props }` ?

src/ContextProvider/index.js
```diff
@@ -5,28 +5,28 @@ import { Component, PropTypes, Children } from "react"
 export default class StatinamicContext extends Component {
   props: {
     collection: StatinamicCollection,
-    layouts: Object,
+    layouts: Object, // deprecated
```
thangngoc89: I know this is not the scope of this PR but flowtype props and propTypes is not the same here.

collection, layouts and metadata is required in flowtype.

thangngoc89:
Awesome UX with topbar and loading spinner

src/ContextProvider/index.js
MoOx: You are right. But without collection or layout, we will have issue. Will check this before removing proptypes

CHANGELOG.md
MoOx: I like to prepare myself for functional component (stateless or [stateful](https://github.com/reactjs/react-future/blob/ba5a275d08451756137559714d79f15055b9a14e/07%20-%20Returning%20State/01%20-%20Stateful%20Functions.js)) :D

MoOx: So now I like to extract props and state has const :)

thangngoc89: Thanks for useful link :)

PR #375, opened 2016-04-11
Remove mkdirp from dependencies

thangngoc89:


MoOx:
What about instead ditching fs-extra, which is pretty big dep, that is (before this PR) used twice, for doing a rimraf/mkdirp and a copy -r. We can replace by cpr ? (since rimraf his already here anyway - dep of fs-extra)...

thangngoc89:
> that is (before this PR) used twice

I'm not sure. But imho, removing fs-extra doesn't make a different. 

- fs-promise depends on fs-extra (and we use many methods of it). 
- fs-extra depends on mkdirp (which this PR trying to remove)

So this only make our package.json cleaner. No more or less dependencies to download.

MoOx:
We can replace fs-promise by pify + fs. fs-extra rely on grateful-fs which is another dep.

thangngoc89:
> replace fs-promise by pify + fs

Good idea. I already do this partially in #343 

So I'll send another PR to get rid of fs-extra and fs-promise after #343 landed

PR #376, opened 2016-04-11
Remove AVA serial script

thangngoc89:
I added a fail test in this PR so unit tests should fail. :/


thangngoc89:
@MoOx I'm not sure why I'm throwing on main process but the process doesn't exit. Can you take a look ?

MoOx:
Use setTimeout to throw when in a promise

thangngoc89:
> Use setTimeout to throw when in a promise

Nice catch. But I prefer setImmediate. It's cleaner.

thangngoc89:
It doesn't fail when run in serial mode 

https://travis-ci.org/MoOx/statinamic/jobs/122192604

ben-eb:
Heh, nice one. :+1:

thangngoc89:
Craps. Unit tests fail on Windows all the time without us knowing about it.

Shame :( 

I saw this error message before when enabling node 4 on Appveyor

MoOx:
If AVA is so annoying and does not fix perf issue soon, I am going to use tape again. We should not loose time on this shit :'(
Or we can get rid of node 4 for now, until they fix their issues.

thangngoc89:
Yes. I'm losing so much time on AVA and node 4. I saw the perf issue was assigned. Maybe please wait a little longer. I'm too lazy to learn another test runner

MoOx:
FYI, AVA is highly inspired by [tape](https://github.com/substack/tape).
Anyway, I think we should remove this way to run AVA tests, and skip node 4 until they fix https://github.com/sindresorhus/ava/issues/78

thangngoc89:
Haha. looks like AVA 0.14 has some performance improvements. All tests passed

thangngoc89:
This blocks other PR so I merge this :)

PR #386, opened 2016-04-12
Add: meta generator tag in HTML result

### LINKED ISSUE #361: Add meta generator tag
Everyone is doing this so I think we should do this too

For example:

```html
<meta name="generator" content="Statinamic 0.9.2 https://github.com/MoOx/statinamic" />
```

thangngoc89:
Close #361


MoOx:
Maybe we should use react-helmet in the boilerplate so people can remove this? (and simpler to maintain).

thangngoc89:
Yes. This make the code more complicated but as I can see from others site generators, there is no way to change this line unless you change the core :D

thangngoc89:
Btw, it will increase JS bundle size since we need to require package.json of Statinamic for information (which is big)

MoOx:
We should inject some var via a private wepack defineplugin instance. We receive users webpack config so should be easy to hide this and document injected vars.

PR #392, opened 2016-04-13
Fix: dev server show wrong log message about duplicate port

### LINKED ISSUE #391: phenomic:builder:server Port 3000 is not available. Using port 3000 instead. +0ms
Reported by @magsout. Maybe related to the string/int thing.

thangngoc89:
- Using number type for devPort right now.

Close #391


MoOx:
Can we keep old behavior as well to avoid a breaking change? Or maybe just support string or number and parseInt in a config specific routine (like for complicated option)?
This is to avoid calling this a change, since currently, it's not a fix, but a change :)

thangngoc89:
Oops. Totally forgot about backward compat. Does yargs support validate multiple types ?

MoOx:
Not sure, but the best thing to do is to remove yargs validation and use a custom check.

thangngoc89:
Never mind. It's a way to type definition from us. Will normalize everything in configurator

thangngoc89:
PR updated. Nice ES6 Math.trunc

src/configurator/validators/devPort.js
```diff
@@ -0,0 +1,15 @@
...
+  { config, errors } : { config: PhenomicConfig, errors: Array<string> }
+): void {
+  const port = Math.trunc(config.devPort)
+
+  if (port > 0) {
+    config.devPort = port
+  }
+  else {
+    errors.push(
+      "devPort can be a string or a number and must be a legal port"
```
MoOx: We should not explicitly allow string (at least in the message).

```
`devPort` must be a legal http port number"
```

thangngoc89:
PR updated.

MoOx:
No test, but I will trust you on this :/

PR #395, opened 2016-04-15
Added: (boilerplate) more layouts and components so new users have more...

MoOx:
…example to show list of entries.

screen shot 2016-04-15 at 07 21 26

No sure why, for the first page (server side rendered), I got a react warning:

```
Warning: Failed propType: Invalid prop `date` of type `date` supplied to `PagePreview`, expected `string`. Check the render method of `PagesList`.
```

No idea why I get a "date" for the date field. gray-matter is not converting date (I checked the code), and client side, clearly have string, so no fucking idea.

### DIFF
--- a/boilerplate/web_modules/layouts/Homepage/index.js
+++ b/boilerplate/web_modules/layouts/Homepage/index.js
@@ -1,0 +1,29 @@
+import React, { Component, PropTypes } from "react"
+import enhanceCollection from "phenomic/lib/enhance-collection"
+
+import Page from "../Page"
+import PagesList from "../../PagesList"
+
+const numberOfLatestPosts = 6
+
+export default class Homepage extends Component {
+  static contextTypes = {
+    collection: PropTypes.array.isRequired,
+  }
+
+  render() {
+    const latestPosts = enhanceCollection(this.context.collection, {
+      filter: { layout: "Post" },
+      sort: "date",
+      reverse: true,
+    })
+    .slice(0, numberOfLatestPosts)
+
+    return (
+      <Page { ...this.props}>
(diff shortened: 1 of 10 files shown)


boilerplate/web_modules/layouts/Homepage/index.js
```diff
@@ -0,0 +1,29 @@
...
+    collection: PropTypes.array.isRequired,
+  }
+
+  render() {
+    const latestPosts = enhanceCollection(this.context.collection, {
+      filter: { layout: "Post" },
+      sort: "date",
+      reverse: true,
+    })
+    .slice(0, numberOfLatestPosts)
```
thangngoc89: enhanceCollection has a limit option for this job.

Maybe we can extend it to support pagination

MoOx: Pagination will be another thing, later :)
Will use limit :)

thangngoc89:
I ran into this before. Can't figure out what's going on.

thangngoc89:
Not sure this is a bug or a feature. This came from gray-matter's dep js-yaml. It will parse date automatically. See https://github.com/dplepage/js-yaml/blob/5f058c3a78fd1a1bd36fd2cdc2c3a8c45ab53fd1/lib/js-yaml/type/timestamp.js#L79

The data which is sent to the client was JSON.stringify so it's a string

MoOx:
I tried to assume it's a date, but somewhere else (client side), dates are string, so not consistent with server cache, which have real dates. We should normalize this. Opening an issue.

thangngoc89:
Could we merge this and deal with the auto parsing date later ?

PR #396, opened 2016-04-15
Internal: improve release process by pushing a master up to date

MoOx:
Currently, I always forget to update master branch, and master generate
the docs via travis. So this “trick” should help to avoid this
oversight.


thangngoc89:
Are you sure about `git rebase next` ?

I think it must be `git merge next`

MoOx:
I am using rebase for a few versions.
Since nothing has changed on master, it just add commits. Merge will create a merge commit, and merge commit sucks. No? (I am not a git pro)

thangngoc89:
> Merge will create a merge commit, and merge commit sucks

No. The merge commit thing is the Github shit. (I do this all the time, but you should double check this)

Using rebase can change the commit hash if we aren't careful

MoOx:
Ok, will try by hand for the next release and will see.

MoOx:
I used merge and it is good.

PR #402, opened 2016-04-15
Internal: Speed up lint process by running in concurrently

thangngoc89:
Before: 23s
After: 14s 

Should I apply this change to boilerplate and docs as well ?


MoOx:
> Should I apply this change to boilerplate and docs as well ?

Sure, even if this will be less noticeable, since flow is probably the biggest process here.

thangngoc89:
> since flow is probably the biggest process here.

Not true.eslint is the lowest.

MoOx:
Did you try https://www.npmjs.com/package/npm-run-parallel ?

thangngoc89:
> Did you try https://www.npmjs.com/package/npm-run-parallel ?

concurrently is better. Since doesn't fail if 1 process fail. 

But I'm rethink about this. Their is no color in the output. :(

thangngoc89:
PR updated.

MoOx:
Everything should fail if one fail.

MoOx:
What about npm-run-all ?

thangngoc89:
> Everything should fail if one fail.

I don't like this. This is a linting process, I prefer get all the errors and fix them all.

MoOx:
See https://github.com/MoOx/eslint-config-i-am-meticulous/commit/45119fd78944c9cc95a64b9cc163f2ebad4dc9af

MoOx:
Anyway, process should exit with the correct code if one fail.

thangngoc89:
Woa. npm-run-all is cleaner. I'll go for it.

thangngoc89:
PR updated.

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 ## Boilerplate
 
+- Added: Run lint process in parallel with [concurrently](https://www.npmjs.com/package/npm-run-all)
```
MoOx: changelog has to be updated :)

package.json
```diff
@@ -167,7 +169,7 @@
     "lint:js:flow": "babel-node scripts/flow-check.js",
     "lint:js": "npm run lint:js:eslint && npm run lint:js:flow",
     "lint:css": "stylelint \"+(boilerplate|docs)/web_modules/**/*.css\"",
-    "lint": "npm run lint:js && npm run lint:css",
+    "lint": "npm-run-all --parallel lint:css lint:js:*",
```
MoOx: lint:* does not work ?

CHANGELOG.md
thangngoc89: Haha. Stupid mistake.

package.json
thangngoc89: It will run `lint:js:flow && lint:js:eslint && lint:js ` => flow and eslint are run twice

MoOx: so we can remove lint:js, and make the code simpler?

MoOx: I tried this on another project

```
    "lint:js:eslint": "eslint --ignore-path .gitignore --fix .",
    "lint:js:flow": "flow src",
```

Got

```
ERROR: Task not found: "lint:*"
```

With 

```
    "lint:js:eslint": "eslint --ignore-path .gitignore --fix .",
    "lint:js:flow": "flow src",
    "lint:js": "npm-run-all --parallel lint:js:*",
```

Not duplicate output. So are you sure?

thangngoc89: Sorry. I was wrong. It runs `lint:js && lint:css`. It treats the colon as sperator so you will need to do this: "npm-run-all lint:**"

MoOx:
FYI https://github.com/eslint/eslint/issues/5893

thangngoc89:
:+1: That's very interesting

MoOx:
Now this take around 5 to 7s.

PR #403, opened 2016-04-15
Normalize all data to JSON compaitable types

thangngoc89:


src/content-loader/index.js
```diff
@@ -31,11 +31,15 @@ module.exports = function(input) {
   const defaultHead = query.defaultHead
   const parsed = frontMatterParser(input)
 
+  // Normalize all data to JSON compaitable types
+  // See https://github.com/MoOx/phenomic/issues/397
+  const parsedData = JSON.parse(JSON.stringify(parsed.data))
```
MoOx: This will slow the process. I would prefer another solution.

MoOx: And what if people expect js-yaml to do his job?

MoOx: Maybe we should not export as json on the client side, since that not just json.

thangngoc89: > This will slow the process. I would prefer another solution.

I agree.

> Maybe we should not export as json on the client side, since that not just json.

This is possible if we remove server side rendering, babel-plugin-webpack loader. I'm waiting for this.

thangngoc89:
Close. Need to find a way do it more efficiently

PR #404, opened 2016-04-15
Docs: Enable long term caching

thangngoc89:
`phenomic.io` has a CDN now so it makes sense to do long term caching.


MoOx:
Not a fan since react and friends are getting update pretty often.

thangngoc89:
It's not that often. It's better then new hash every build. And I only define a few packages to cache.

MoOx:
Ok to try this.
Maybe we could even add all lib referenced from package.json ( deps , not devdeps)

PR #408, opened 2016-04-19
Add some polyfill

### LINKED ISSUE #407: PageContainer component doesn't work on IE10
We need to add `import "babel-polyfill";` to src/PageContainer/components.js because [collection.find](https://github.com/MoOx/phenomic/blob/next/src/PageContainer/component.js#L49) not works in IE10. With polyfill it works :)

thangngoc89:
- Array.prototype.find (IE10)
- String.prototype.endsWith (Safari 8)

Close #407

I disabled eslint in that file. I want a clean copy-paste

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
-
+- Added: Polyfill for non-ES6 browser compatibility.
+  ([#407](https://github.com/MoOx/phenomic/issues/407))
 - Added: Initial support for Service Worker
   ([#343](https://github.com/MoOx/phenomic/pull/343))
 - Changed: Remove the ability to define a custom globby pattern via appcache option.
--- a/src/client/index.js
+++ b/src/client/index.js
@@ -1,5 +1,6 @@
 // @flow
 // App
+import "./polyfill"
 import React from "react"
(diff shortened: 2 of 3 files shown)


MoOx:
I think I would prefer to include babel-polyfill to avoid having to maintain a list of polyfill. What is the problem with babel-polyfill?

thangngoc89:
Babel-polyfill is big and slow down the initial load. If you go for babel-polyfill,  please include it in boilerplate so that I can remove it. I prefer cdn polyfill,  it smarter and much faster (fastly cdn)

MoOx:
So why not including the cdn polyfill?
Otherwise, we can remove endsWith and find and other problematic code from client code base. Maybe it's even simpler?

thangngoc89:
> So why not including the cdn polyfill?

Well. we can't. #348 must be resolved first.

>  endsWith

No. it's in react-router (or something else I don't remember, but not in phenomic codebase)

> find

Yes. Just use a ponyfill instead of polyfill. 

But by doing so, we are basiclly maintaining a list of polyfill for ES6 features. 

I'll resolve #348 and add cdn polyfill to boilerplate.

PR #409, opened 2016-04-19
Internal: Remove npm install npm@latest

MoOx:
Useless now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,9 @@ cache:
     - docs/node_modules
 
 before_install:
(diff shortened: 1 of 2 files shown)


thangngoc89:
Node 4 ? We do need npm@3 for it.

MoOx:
So we might use a smarter command to avoid this install on node 5...

thangngoc89:
I don't see any failure in the build. May be because of the caching.

thangngoc89:
LGTM

PR #410, opened 2016-04-19
Wrong 'sw-register.js' path when website is not at domain's root

thangngoc89:


MoOx:
We need a test to avoid a regression in the future.

thangngoc89:
I don't know how to test that file. Do you ?

MoOx:
Add a new use case here with SW on and a base url https://github.com/MoOx/phenomic/blob/next/src/static/to-html/__tests__/url-as-html.js

thangngoc89:
You're right. :+1: I added a test case in #348 . So I'll close this continue overthere.

MoOx:
Maybe we should fix this issue independently? To avoid the pain of relying on a major breaking change?

thangngoc89:
Actually, this PR is unneeded. The baseUrl was added later in the same file. And I didn't notice about that.

MoOx:
So false alarm :)

Issue #412, opened 2016-04-19
Add a private definePlugin instance to inject PHENOMIC_* constantes


MoOx:
For #386

thangngoc89:
Maybe remove PHENOMIC_PATHNAME on the way too ?

MoOx:
Will reconsider this. Might be exposed by core (and so removed from boilerplate) if this constant is needed.

Issue #363, opened 2016-04-04
Validate CLI args


MoOx:
For now, we cannot validate CLI args properly, since babel-plugin-webpack-loaders will add some (when spawning webpack).
Might be revisited after we ditched this plugin, see #301.

Issue #40, opened 2015-10-26
Consider replacing redux in favor of react context


MoOx:
Since we don't do a lot of modifications (ie: no modification at all for now), we might consider replacing redux usage by react context (that are now documented http://facebook.github.io/react/docs/context.html).
That should be clearly enough to just "pass" some data (collection and page components references) that are declared during app initialization.

It don't really see any downside from the current usage.

Any thoughts? poke @bloodyowl

bloodyowl:
I'd really say it depends on what you try to achieve. statinamic already seems opinionated on some parts, and I see it more like a boilerplate than a library. if you want to provide a way to manage data across components whatever the tree structure is in this boilerplate, then I'd say keep redux. If you want it to be more of a library, keep it minimal, and just use your own provider and decorator.

MoOx:
Statinamic is a collection of lib + a boilerplate generator + docs: you start with a boilerplate that use statinamic libs to help you generate a website with react and webpack. So Statinamic is opininated because of these 2 things (react and webpack). Except that nothing is mandatory (no required structures etc).
You can use the webpack config you want, your own loader, and any npm module you want in your components.

There is a minimal boilerplate that is required to allow some flexibility (choose your styles flavor (inline, css modules, sass...) and I am currently writing a helper to generate the minimal required boilerplate.

The idea is to allow to create a website very easily with one command `$ statinamic setup`: this will generate the boilerplate that use Statinamic libs. The only things that Statinamic uses are:
- a registry to access page data
- a registry to get available pages layouts (component references) that can be used from md header

Not a lot of things, no user interaction. That's what I was thinking about using react context instead of redux. Currently there is already a context provider to share some metadata about your project (eg: package.json to get homepage url, stuff like that) and I was thinking about using it for all the shared (read only - or sort of) data. The only interaction is: you click on a link, if page data are not in the store, we fetch a json to get those.

thangngoc89:
I use redux in almost every project so using redux is one the reasons why I use statinamic.

Also with redux, I can just add reducers that listen for default action of statinamic. Some example use cases: show a loading indicator when getting new page's info. show a error message when getting new page's info failed, ...

In the other hand, keeping redux would make statinamic too opinionated.

MoOx:
> show a loading indicator when getting new page's info. show a error message when getting new page's info failed, ...

I think thoses things should be possible without redux (and more easy to implement without using redux).

I will try to avoid redux and maybe will consider removing it.

[...]

PR #422, opened 2016-04-27
Refacto perf

### LINKED ISSUE #301: Performance improvement

MoOx:
Best to review with [?w=1](https://github.com/MoOx/phenomic/pull/422/files?w=1)

Closes #301 

Since there is a pretty considerable change in the `scripts`, that's probably the time to handle those issue before cutting a release:

- generate all phenomic requirements #412
- remove redux and the store #40
- validate CLI args #363


thangngoc89:
:+1: Looks fantastic on a quick run. I'll take a closer look tonight

MoOx:
The issue on Appveyor seems related to a babel-register issue. Will take a look later.

docs/content/docs/faq.md
```diff
@@ -2,6 +2,20 @@
...
 
+## ``[BABEL] Note: The code generator has deoptimised the styling of ...``
+
+_This is not an issue. Just a simple information._
+
+When babel compiles some code, it try to offer a readable output, but when files
+become big (>100KB), babel considers (by default) that it's not useful to keep
+a readable output.
+If you are annoyed by this message, or if you don't care about readability,
+you can just use `compact: true`.
```
thangngoc89: Could you make this default option ?

docs/scripts/phenomic.node.js
```diff
@@ -0,0 +1,13 @@
+import metadata from "../web_modules/app/metadata"
+import routes from "../web_modules/app/routes"
+import store from "../web_modules/app/store"
+
+import phenomicStatic from "phenomic/lib/static"
+
+module.exports = (options) =>
```
thangngoc89: Why not `export default` here ?

src/builder/cleanup.js
```diff
@@ -0,0 +1,10 @@
+import { join } from "path"
+import { unlinkSync } from "fs"
+
+export default (wd, fileToKeep, assets) => {
+  assets
+    .filter((file) => file !== fileToKeep)
+    .forEach((file) => {
+      unlinkSync(join(wd, file))
```
thangngoc89: How about using `unlink` promise and `Promise.all` ? I think it would help the perf

src/bin/index.js
```diff
@@ -1,8 +1,18 @@
 #!/usr/bin/env node
 
+if (!process.env.DEBUG) {
+  process.env.DEBUG = "phenomic:*"
+}
+
 // still in src/, handy to play
 if (__filename.indexOf("phenomic/src") > 1) {
```
thangngoc89: I think this is why build failed on Appveyor. This requires the phenomic source folder must be named `phenomic`. Maybe you can try with `indexOf("src/bin/index.js")`

docs/scripts/phenomic.node.js
MoOx: node compat. This script is consumed via node. But I will try to use export default and adjust in the core usage

docs/content/docs/faq.md
MoOx: Maybe. But I don't think "true" or "false" are better. Auto is a good choice. But why not forcing true by default.

src/builder/cleanup.js
MoOx: I am lazy, feel free to adjust this ;)

src/bin/index.js
MoOx: This is unchanged, but maybe that's the reason since I changed something in another place. Will try something else.

thangngoc89: Yup. I noticed that. But I didn't change my local folder (stills statinamic`) and it throws the extract error on Appveyor when I build docs.

docs/content/docs/faq.md
thangngoc89: yeah. User can always change this. But no warnings by default is better

src/builder/cleanup.js
thangngoc89: I'll leave it there for now. This will be a lot easier if we use async/await for `src/builder/index.js`

thangngoc89: https://ci.appveyor.com/project/MoOx/phenomic/build/1580#L2452

Path on windows is different from linux. ('/' vs '\')

src/bin/index.js
thangngoc89: https://ci.appveyor.com/project/MoOx/phenomic/build/1580#L2452

Path on windows is different from linux. ('/' vs '\')

thangngoc89:
I fixed Windows compat issue.

src/bin/index.js
```diff
@@ -5,7 +5,10 @@ if (!process.env.DEBUG) {
 }
 
 // still in src/, handy to play
-if (__filename.indexOf("phenomic/src") > 1) {
+if (
+  __filename.endsWith("src/bin/index.js") ||
+  __filename.endsWith("src\\bin\\index.js")
```
MoOx: can you use node path.sep instead ? Or even better, join("src", "bin", "index.js").

thangngoc89: I updated it with `path.join`

thangngoc89:
LGTM

kuzvac:
If you remove redux/store we can use it like regular redux/store or how?

thangngoc89:
@kuzvac you can add it back like you would do in any React project.

MoOx:
Yeah sure, that's planned ;)

kuzvac:
Ok, thank you all :)

PR #427, opened 2016-04-27
Internal: Replace Node 5 by Node 6 on CI

MoOx:
Blocked by : 

- [x] https://github.com/appveyor/ci/issues/770
- [x] https://github.com/tschaub/mock-fs/issues/106


thangngoc89:
Another blocker
https://github.com/tschaub/mock-fs/issues/106

tschaub:
If you upgrade to mock-fs@3.9.0, you should be able to polish this off.

MoOx:
Now it's seems that linked webpack is causing the issue (the error comes from the fact that 2 webpacks deps are used, if my memory is correct, that why we use symlink, due to the fact that our test are run in a subfolder, and not a parent folder, like in real life).

thangngoc89:
Yeah. I know. I suppose because of the way node 6 resolve dependencies.

PR #436, opened 2016-05-04
Pin nyc at 6.4.1

thangngoc89:
See https://github.com/bcoe/nyc/issues/246


MoOx:
Ref https://github.com/bcoe/nyc/issues/246

MoOx:
lgtm

Issue #442, opened 2016-05-05
Performance, round 2


MoOx:
Now that perf are back to "normal" (limited by webpack speed), we need to find tricks to make webpack build faster
Maybe

- https://webpack.github.io/docs/list-of-plugins.html#dllplugin
- https://github.com/amireh/happypack
- https://github.com/markfinger/unfort

An interesting issue: https://github.com/webpack/webpack/issues/250

Also: http://www.slideshare.net/trueter/how-to-make-your-webpack-builds-10x-faster

thangngoc89:
I tried happypack with phenomic before. Only gain 1-2s. I tried it again soon

MoOx:
Just an idea: find a way to detect if there is any file change (or not) to decide if we need to compile via webpack, or just use previously compiled files. This way we could skip webpack build step (eg: if no files in web_modules has been changed + node_modules tree untouched). This may be convenient when you only write content.

thangngoc89:
It sounds like a good idea. But I don't know how hard would it be to implementing such a feature like that.

MoOx:
Another idea: think about removing webpack (or making it optional, eg: if you just have JS(X), you can just use babel and a "simpler"(faster?) tool like browserify). That can definitely take a thorn out of our side.

MoOx:
The webpack dll thing should help a lot. Will try to work on that soon as it should not be complicated.

Also https://github.com/FrendEr/webpack-optimize-example

MoOx:
https://www.npmjs.com/package/hard-source-webpack-plugin

https://github.com/webpack/webpack/issues/250#issuecomment-240643985

> I haven't been able to test on a large project yet, but testing on two smaller personal projects I've seen times for a ~5s project drop down to ~500ms and a ~8s project drop down to ~1s. Those projects respectively have ~600 and ~400 modules.

OMG. We have to try this.

thangngoc89:
There is 1 place we can improve: disable Uglify plugin for node build (whether production or not)

thangngoc89:
I gave hard-source-webpack-plugin a try and it's ridiculous fast 

```
D:\web\phenomic\docs (hard-source-wp)                             
λ npm run build                                                   
                                                                  
> Phenomic@ build D:\web\phenomic\docs                            
> phenomic build                                                  
                                                                  
  phenomic:builder ✓ Static build: assets copy completed +0ms     
  phenomic:builder ✓ Static build: client build completed +27s    
[...]

PR #454, opened 2016-05-07
Use a temp folder for node build

### LINKED ISSUE #439: phenomic.node build might create some files in scripts/_
This can happen if you have some files loaded using file-loader.
A cleanup is made for the extracted css (to only keep phenomic.node.bundle.js), so we should add this to the cleanup.

thangngoc89:
Close #439 

There should be an integration test for this. But I don't want to make boilerplate's webpack config more complicated. Do you have any suggestions ?


MoOx:
What do you think about moving the builded content into node_modules/.cache/phenomic or something like that (AVA does this I think). That would allow us to keep it, and that will be nice with https://github.com/MoOx/phenomic/issues/442#issuecomment-217252645 in mind

thangngoc89:
It's totally fine with be. Need to test to see if it is doable.

thangngoc89:
Do we have any guarantee in codebase that `process.cwd` is the parent folder that contains `node_modules` folder ?

MoOx:
process.cwd should not be used except for test maybe. We have config.cwd in our codebase.

MoOx:
@thangngoc89 do you plan to update this PR according to my first comment?

thangngoc89:
@MoOx could you help me with this? I don't have much time for Phenomic atm :/

MoOx:
Sure. I will handle this.

PR #492, opened 2016-05-19
Creating uri.js for tests for URI joining

ezralalonde:
The join-uri package has some weird semantics, acting both as a unix/windows filepath fixer, as well as (allegedly) for fixing extra slashes in URIs. Right now specifying a scheme in a URI is broken.

I added some (mostly failing) tests for join-uri, in advance of a larger refactor. I hope it helps.

I might recommend that someone with more knowledge of the situation identify the separate scenarios where join-uri is used,:

1. Where join-uri is used for joining unix or windows paths.
2. Where join-uri is used for actual URIs.

It might make sense to use a separate utility for each, as they are separate things, and (it seems) good tools might exist for one or the other, but not both.

I understand that there's some mixing since Phenomic URLs are based on filesystem paths, so it gets a little weird. The windows support thing makes it less clean for URIs (normally a "hybrid" approach using filepath as a subset of URI would be fine. URI as a subset of filepath, as is now the case, is somewhat buggy and incomplete). The current implementation of join-uri supports filesystem paths primarily.

I didn't want to make a larger change, and end up breaking something, since I would end up touching about 20 files. Let me know if there's any way that I can help.


thangngoc89:
Yeah. You're absolutely right. There should be a seperate utility for joining url. join-uri was created for filepath in the first place.

MoOx:
Sorry guys but joinUri is only used to join uri or url. When you see "fileName", joinUri is used with baseurl, so it's definitely to join uri, and nothing else.

ezralalonde:
I guess I was confused by the windows-specific code for handling "\\" in paths, which would appear to have nothing with URIs, but a lot to do with filepaths.

Is it the case that there's a few simple cases where we can `normalizePath` (or whatever you want to call it, to get rid of "\\" for windows users) and treat the rest of them as pure URIs?

MoOx:
Since it's one line of code, not sure it worst creating a function for this. Doesn't seems like a big deal to me. Browsers for example, also convert \ to /.
Do you need some help to define the minimal flow interface for url-join module?

ezralalonde:
I wholeheartedly agree that, if `join-url` (or another package) is the way forward that keeping `join-uri` around is not necesary.

The last commit an attempt to show what breaks when the current behaviour changes. There are several places that appear to depend (at least according to their tests) on joinUri never resulting in two leading slashes.

For example, the meaning `<script src="X"></script>` for X="/test.js" and X="//test.js" is very different, and using `join-url`, absent other (perhaps significant) refactor would change this behaviour in places. 

My confidence that I'm understanding all of the implications of the change is *quite* low, and it's entirely possible that I'm missing or misunderstanding something.

MoOx:
Current usage of joinUri does not include (except the buggy place for the og:url) any hostname right?
Also, protocol relative uris technique is now an anti-pattern (see http://www.paulirish.com/2010/the-protocol-relative-url/) and so I am not sure if we should allow that.
What do you propose as a viable solution?
Maybe if you want we can
- we can create a pathToUri() method (mainly for windows) to ensure filename used to build url use slashes
- use this method in the few places where filenames are used in joinUri call
- remove internal joinUri call
- expose url-join as joinUri.

The last point is (imo) important for many reasons:
- this kind of very small utilities are must have for static generator (eg: the og:url)
- backward compat. About this point, we might keep the exposed method for a while, and remove it later if you think exposing a dep is a bad practice (I remember grunt stopped to expose lodash as a utility ).

Thoughts?

ezralalonde:
I agree that it's beneficial to expose `joinUri` or equivalent, and that it behaves as expected. I don't tend to use relative protocols personally, but I think that it breaks user expectations to have it randomly disallowed.

In my case, I the only time I use `joinUri` is when I need canonical URL for e.g. "og:url". In looking at what other people were doing I commonly saw a pattern of `pkg.homepage + __url` or similar instead. I currently have `urljoin(process.env.PHENOMIC_USER_URL, __url)` in my project.

I agree that you don't want to break backward compatibility.

I also agree with your approach of having two functions. It's a separation that I was imagining in the original post, it was just not clear to me where that boundary was, or whether it was actually paths or URIs were needed internally.

If it is true that the only uses of `join-uri` are

1. Internally (strictly for paths and relative URIs), and
2. In templates/components (where anything is possible)

then maybe the simplest way is to

+ Rename `join-uri` to `pathToUri` or `normalizePath` or whatever, and use it only internally
+ Expose `url-join` or similar as `joinUri` for use in templates/components/etc.

It may not be necessary to draw a distinction between Path/URI internally, as long as Phenomic never internally generates or references absolute URIs. I don't see why it should? You could debate whether it was aesthetically superior to use `pathToUri` in a few places, and `joinUri` in others, but it won't make a functional difference?

MoOx:
>  I commonly saw a pattern of pkg.homepage + __url

I was in the boilerplate. Your method is definitely cleaner.

---

>
    - Rename join-uri to pathToUri or normalizePath or whatever, and use it only internally
    - Expose url-join or similar as joinUri for use in templates/components/etc.

This solution looks good to me.

> Phenomic never internally generates or references absolute URIs

Phenomic only generate uris without domain, but uris starts with basepath (in most case /).

> You could debate whether it was aesthetically superior to use pathToUri in a few places, and joinUri in others, but it won't make a functional difference?

Yeah. pathToUri lgtm. And if needed we can still use url-join + pathToUri.

PR #496, opened 2016-05-20
Node 6 support

MoOx:


kuzvac:
Maybe we need notify user on node v6, that phenomic currently not work properly on v6?  Maybe add rule inside package.json?

thangngoc89:
@kuzvac phenomic can work on node 6 properly. I tested it. This PR is about the internal structure to support node 6 on CI build

MoOx:
I tried this morning with fresh node_modules and v6, didn't get any issue so I will try to update this branch and see what happen, maybe now everything has been fixed in deps.

kuzvac:
Hum, yes, its works. I think it's problem on my side. Excuse me :)

thangngoc89:
So it just works without any modification ? Weird

MoOx:
Now, previously it did not. I cleaned cache so we get latest deps patches.

thangngoc89:
Yeah. I think may be there is some patches from upstream

Issue #489, opened 2016-05-19
join-uri problem?


ezralalonde:
First I'd like to thank you for all the work that has been put into Phenomic. It's a fine piece of software that I'm really enjoying using, and I appreciate all the effort that has been put into making it.

Today I noticed that my "og:url" meta was incorrect, and discovered that the cause a possible fault in `join-uri`: it's removing all instances of "//", including those in schemes, which I believe is an issue.

In my case, I expected `https://some.site`, I get `https:/some.site`. Is this expected? If so, what is the scheme-safe alternative?

    joinUri("https://some.site/", "page")
    >>> https:/some.site/page  // I thought I'd get https://some.site/page

In my local version (only where I set "og:url"), I've replaced this call to `join-uri` with one to `https://www.npmjs.com/package/url-join`, which works as expected.

I originally experienced this issue on 0.12.0, and have since upgraded to 0.12.1, and still see it.

I apologize if this issue is already known, or if this is the expected behaviour of `join-uri`.

I'm happy to submit a pull request swapping calls to the homegrown `join-uri` with the `url-join` package referenced above, if it makes sense, and there's any interest in it.

Thanks again for all your great work!

Cheers,
Ezra.

MoOx:
This is definitely a bug.
Feel free to make a PR to replace internal utility by the one you mentioned, otherwise I will end up doing it.
Just be sure to continue to expose it in phenomic index like  it was before so we can release this change as a patch. Another is planned for #469.

ezralalonde:
Cool. I'll submit a pull request this afternoon.

Thanks in advance for your patience in dealing with me…

Cheers,
Ezra.

MoOx:
No problem. At the stage of this project, every feedback and discussions are welcome :)

ezralalonde:
Turns out the issue, internally, is a little hairy, since there's not curretly a clear distinction between when it is appropriate to treat things as filesystem paths, and when they are URIs. `join-uri` is, as far as I can tell, is being used for both.

I just ended up writing some tests, since I don't want to break any internals.

PR #500, opened 2016-05-25
`join-uri` now `path-to-uri`.

ezralalonde:
+ Renamed `join-uri` to `path-to-uri`
+ Made `path-to-uri` internal-only
+ Exported `url-join` publicly in place of `join-uri`
+ Updated docs and boilerplate

Discussion at 
+ https://github.com/MoOx/phenomic/issues/489


thangngoc89:
Sorry for late comment. But naming the internal utility path-to-uri doesn't make sense to me. As I can see from the original issue, join-uri joins urls start with protocol incorrectly. Why don't we join treat it as a special case for join-uri ?

ezralalonde:
All `join-uri` does is join arguments with a '/' and strip out extra slashes and transform '\' to '/' in windows paths, not true URI joining. The name `path-to-uri` is meant to show that, rather than applying to all URIs, it really only works for filepaths that need to be represented as URIs. In paths a `//` is never valid, so the resulting URI should never have a `//` in it. Similarly, real URIs should encode '\' as '%5C' (or whatever), which won't work for our case of getting windows paths to URIs, obviously.

I agree that the name is maybe a little weird, but names are hard. I don't think it's captured any better by `join-uri`, as this makes people thinking that you're *actually joining URIs*. Maybe `join-path-as-uri` or some-such is a better alternative, I don't really care that much, personally.

There's some additional discussion at https://github.com/MoOx/phenomic/pull/492 that touches on this a little bit.

The internals, as far as I know, deal in relative "URIs" only, and have no need to deal with the mess of real URIs. This makes perfect sense. There's also a function that works perfectly for this need: `join-uri`/`path-to-uri`/whatever-the-best-name-is.

If you "fix" the implementation of the current `join-uri`/`path-to-uri` function to support the protocol, then you end up breaking as much as you fix, as there are a number of places where stripping two leading slashes is the expected/required behaviour, since they are coming from windows/posix paths, and not "real" URIs.

It is currently, I believe, easiest/cleanest to have two functions:

1. One does what phenomic wants internally (basically fixing paths into usable URI chunks) for paths
2. One that does full-blown URI joining in all its glory, for when you actually need URIs.

Mixing these functions not only hard to get right but, one could argue, *wrong*, since they are different things. For example, an actual URI joiner would escape the backslashes, not replace them with a forward-slash, as `path-to-uri` *must* do to work as advertised on Windows. At best, I think, we could just make the URI joining wrong in increasingly subtle ways, but it's going to be a lot of effort for (in my opinion) very little gain.

The only case for 2 so far is "og:url" and whatever else is in "user space". You could argue that phenomic should not export joinUri at all, and that users could simply use `join-url` where they please, but it may have some effects on backward-compatibility and upgrading.

Note that you could definitely isolate the use of `path-to-uri` better than I have, and replace several calls with a real URI joiner if you wished. As it stands, I've done so in the most brute way possible, but it at least promises that nothing breaks from the change.

If there's something I'm not aware of, or my reasoning or implementation (or anything else) is off, please let me know. I could be way off base.

ezralalonde:
PS. The test at https://ci.appveyor.com/project/MoOx/phenomic/build/1825 is, weirdly, failing. It passes locally (on my linux box) and basically asserts that the current observed behaviour (stripping the leading `//`) isn't happening. I blame windows?

thangngoc89:
Yeah. Thanks for the long and great explanation. This is PR is the best way for doing this. Also, when test fails on appveyor. You blame windows. It sucks.

thangngoc89:
I took a look at the failing test. I think maybe this is a bug of nodejs path.join ? Since we do nothing rather than wrapping that library. I'll find a windows box and test it.

MoOx:
The failure looks like it's a node inconsistency.

MoOx:
So we should probably replace // by / ourselves.

ezralalonde:
I believe that I've addressed the offending test/platform difference by forcing the platforms to act the same. Node's "path" is different for the platforms, and likely the cause of error.

This attempt:

1) Manually replaces \\ with /, as we'd been doing, but in all cases instead of just win32
2) Calls `path.join`, as we'd been doing, but forces it to always use the POSIX join (`path.posix.join`), so it's no longer platform-dependent
3) calls `normalize` (`path.posix.normalize`) on the resulting path, thereby de-duping our slashes.

I don't have a Windows box to test on; I apologize if it's still not passing.

MoOx:
Thanks @ezralalonde 
This lgtm this way.
@thangngoc89 thoughts?
If you are ok, I can merge and cut a patch release.

thangngoc89:
LGTM. thanks a lot @ezralalonde

MoOx:
Released as 0.12.3

PR #503, opened 2016-05-27
Internal: Fix regression with path-to-uri introduced in #500

### LINKED ISSUE #502: Regression with 0.12.3: path.posix does not exist in the browser
This breaks everything on the client side.

@ezralalonde any idea how to handle this without path.posix? (I don't have much time atm).

Meanwhile, I tagged 0.12.2 as the latest version.

thangngoc89:
Close #502

### DIFF
--- a/src/_utils/path-to-uri/index.js
+++ b/src/_utils/path-to-uri/index.js
@@ -1,10 +1,12 @@
 // @flow
-import { posix } from "path"
+import { join } from "path"
 
+/**
+ * Normalize path into uri
+ * Join, replace multiple / or \ to single /
+ */
 const pathToUri = function(...args: Array<string>): string {
-  return posix.normalize(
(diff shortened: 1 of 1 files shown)


MoOx:
So here you are just providing the posix behavior by hand? @ezralalonde thoughts?

thangngoc89:
@MoOx Yes. I'm providing posix behavior by hand

ezralalonde:
Sorry for the oversight. *C'est la vie*.

This seems fine to me; I like this version the most so far.

Looking at the difference between `posix.join` and `win32.join`, `win32` has a special case for network/UNC paths, which is what was causing platform-dependent problems.

Having "/" and "\" in the regex, and de-duping before `join` makes both `win32.join` and `posix.join` act the same, as far as I can tell, since the network/UNC case is avoided.

MoOx:
Thanks you guys!

PR #507, opened 2016-06-05
Add Remark plugin to transform gemoji shortcodes into emoji

azat-io:
I think it would be better if the user be able to use [Gemoji shortcodes](http://www.emoji-cheat-sheet.com/) for converting `:smile:` to :smile:

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -59,6 +59,7 @@
     "redux": "^3.0.0",
     "remark": "^4.1.1",
     "remark-autolink-headings": "^3.0.0",
+    "remark-gemoji-to-emoji": "^1.0.0",
     "remark-highlight.js": "^3.0.0",
     "remark-html": "^3.0.0",
     "remark-slug": "^4.1.0",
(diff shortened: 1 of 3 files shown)


MoOx:
Nice idea.
Not sure if this days, it's very useful. I daily use the OS X emoji picker (using cmd+ctrl+space) so I don't really see a direct benefit. Any comments?

azat-io:
I think now many users still use Linux or Windows OS in their laptops (for example like me). And it'll be very helpful for me

thangngoc89:
I'm not sure if this is necessary. If you want this in your website, you can use a custom render. IMHO, the default renderer should be lightweight

MoOx:
@azat-io did you successfully added this via a custom renderer?
Maybe we should expose the default remark plugins list so people can easily inherit it and just add one or 2 plugins?

thangngoc89:
@MoOx that's a good idea since remark's dependencies are defined in phenomic's . If user want to add something, they have to install a bunch of deps (4-5)

MoOx:
Will eventually add that in the docs when #668

PR #514, opened 2016-06-07
Added: (boilerplate) ``*.global.css`` can be just normal CSS (not CSS Modules)

MoOx:
- Remove :global css from css modules
- add some .global.css files
- add docs about that in config and on the docs
- add docs about potential usage of Sass and LESS.


thangngoc89:
LGTM

Issue #451, opened 2016-05-07
Random 404, even on production, when using client side navigation


MoOx:
I just opened https://phenomic.io/ and looked for Usage / Deploy to github and got a 404. This is bad. Maybe related to SW.

thangngoc89:
I saw that too. But only 1 time, I can't reproduce it anymore to find out what causing this problem.

MoOx:
I think the following scenario create this issue:

- I have offline mode with SW
- I open the homepage
- later (after an update of the website), I reopen the homepage: cache is used since our SW say "cache first"
- in this homepage, collection reference to json files are out of date, and if file is not cached already, when I will open a page, I will get a 404 (client side 404).

According to this, we must change our SW policy as we provide json with content hash. Or if people want offline first, we need to remove hashes so url to json (like html url) never changes.

But using by default network first will also "fix" #277

thangngoc89:
Yeah. That might be the scenario here. We clean cached content but doesn't change json files link.

MoOx:
I am thinking about even removing content hash on json. Here is why.

- I load a page "a" (hash "123")
- Another day I load another page "b". The collection in this page indicates that page "a" has changed (hash "456"). Here current cache for "a" is deleted since out of date.
- Another day, I am offline, and goes to page "b" (works, offline cache via sw), then try to access page "a". Hash is "456" and I don't have this page in cache. But I previously had the "123" version. 

It would be nice to have the previous (even if not up to date) page, right?

Like normal url, url for json should not change. Cool uris don't change. Don't you think? Anyway, browser handle cache and when we add an hash, we are trying to do this and cache is a hard problem, we all know that.

You might say that the same can apply for css and js.

thangngoc89:
Adding hashes is mandatory for some cdn to work properly. Removing them is not a good idea. I think that we should find a way to update the current collection in memory so it will reflect the new content changes.

MoOx:
The collection is always up to date, the problem is more that you don't have out of date cached pages as fallback.

MoOx:
Not sure if hashes are mandatory. It's not for html files isn't it?

thangngoc89:
Some cdn caches files forever. So it's mandatory to invalidate those files

MoOx:
So the problem applies for html files as well? More resources about that?

thangngoc89:
I don't know about html files.

More info (I use Amazon Cloudfront as an example): 

[...]

PR #535, opened 2016-06-21
Changed: Service Worker now use network first approach

### LINKED ISSUE #398: Inject sw-register.js into the main js bundle
Since sw-register.js is pretty small, making another http request to get it is costly. We should inject it to the main js bundle instead.

---
Below is the plan by @MoOx 

In config we have `config.webpackConfig` and `config.webpackConfigClient`. We can inject the file we want in `config.webpackConfigClient.entry.phenomic-client` (btw, we may validate that this entry name exist, or at least that there is one).

For the injection we can do: 
- if there is a single entry, just append (entry can be a path OR a array of path)
- if there is multiple entries, and one is called phenomic-client, append to this one.
- if there is multiple entries, and no one is called phenomic-client, throw an error for now (if people ask we might allow the name to be changed, but not for now - or if you want, add a param in config like "entryToInjectScripts", default to phenomic-client, cause maybe we might add more for other features).

### LINKED ISSUE #485: Change default service worker behavior
Currently it's cache first. But it's not relevant for our docs.
A better behavior is network first, cache if offline.
https://github.com/GoogleChrome/sw-toolbox and https://github.com/GoogleChrome/sw-precache can help.

MoOx:
This is to prevent having an outdated collection that will lead to request outdated entries urls that might not exist anymore (and that have not been cached yet, see [#451](https://github.com/MoOx/phenomic/issues/451)).
This is for the better.
By the way, we now rely on [offline-plugin](https://github.com/NekR/offline-plugin) to generate Service Worker and Appcache files. The part that register the Service Worker is now included in the client bundle.

Closes #485
Closes #398


thangngoc89:
Nice work. I didnt realize that we can use offline plugin now. But does the old config still work? I dont see docs updating in this PR

MoOx:
Yeah should work. Look at the webpack config injection part.

I will still double check that it should work has before (except for the network first change) before merging.

MoOx:
@thangngoc89 it seems I didn't finish the job: the config is not really used haha. Will update this tomorrow. Thanks for your review ;)

thangngoc89:
Hey, also could you add a simple docs on how to config those thing? You know, for common cases. I dont wanna read the entire docs of offline-plugin for a small change

MoOx:
I am updating the config format to offer an simple UX.

The config will look like this

```
type PhenomicOfflineConfig = {
  serviceWorker: boolean,
  appcache: boolean | {
    onInstall: boolean,
    afterInstall: boolean,
    onDemand: boolean,
  },
  cachePatterns: {
    onInstall: Array<string>,
    afterInstall: Array<string>,
    onDemand: Array<string>,
    excludes: Array<string>,
  },
}
```

Defaults will likely be

```js
const defaultOfflineConfig = {
  serviceWorker: true,
  appcache: {
    onInstall: true,
    afterInstall: true,
    onDemand: false, // not sure about this. Is it a good compromise?
  },
  cachePatterns: {
    onInstall: [ "index.html", "phenomic.*" ], // minimal appshell, not enough if people use image in the layout
    afterInstall: [ "**" ], // TO DETERMINE, not sure if we add all be default or just **/*.json (but for example, this will miss assets like images...
    onDemand: [ ":rest:" ], // everything that have not been cached or excluded (this is a special keyword for offline-config)
    excludes: [ "**/.*", "**/*.map", "**/*.html" ],
  },
}
```

Thoughts?

I used the terminology of SW since appcache days are counted.
Docs & config validation will be updated accordingly. Hopefully I will finish this important change this week.

thangngoc89:
Woa.This new config stuff is pretty good. I really like it.

MoOx:
PR updated :)

Note: since we now use network first approach, sw is not disabled in development - and the nice part is that you don't wait with a blank screen while webpack is building ;)

thangngoc89:
I dont like the idea of having SW enabled on dev mode. I use localhost:3000 for multiple projects. This can create confusions. (Why do I see a different website after starting dev mode)

thangngoc89:
I'll take a look at the PR soon

MoOx:
I might add an option to disable completly during dev.
I will change the default port after this PR anyway to avoid the problem you mentionned, seems a better thing to do. 3333 ? Or we can choose one "free" in here https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers

thangngoc89:
3333 is nice. Since we have portfinder, port collision is not a problem

CHANGELOG.md
```diff
@@ -1,3 +1,54 @@
+# HEAD
+
+- Removed: ``appcache`` option has now completely being removed in favor of
+  ``offline`` (after being deprecated in 0.11.0). See new API below.
+- Changed: Offline option has been completely rewritten.
+  Service Worker now use network first approach.
```
ben-eb: *now uses a network first approach.*

CHANGELOG.md
```diff
@@ -1,3 +1,54 @@
...
+  This is to prevent having an outdated collection that will lead to request
+  outdated entries urls that might not exist anymore (and that have not been
+  cached yet, see [#451](https://github.com/MoOx/phenomic/issues/451)).
+  This is for the better.
+  By the way, we now rely on
+  [offline-plugin](https://github.com/NekR/offline-plugin)
+  to generate Service Worker and Appcache files.
+  The part that register the Service Worker is now included in the client
```
ben-eb: *registers*

docs/content/docs/advanced/offline-browsing.md
```diff
@@ -1,105 +1,200 @@
...
-Phenomic supports offline browsing **out of the box.**
-
-There are two different technique to enable offline experience for your website:
+There are currently two different technologies to enable offline experience.
 
 ## AppCache
 
-This is the old one, IE 10 (and above) supports it.
+Appcache is the oldest way to offer offline supports but it's a bit brutal since
+you can only choose what to save when the website open.
```
ben-eb: *the website is open.*

docs/content/docs/advanced/offline-browsing.md
```diff
@@ -1,105 +1,200 @@
...
 - [Learn about AppCache](http://www.html5rocks.com/en/tutorials/appcache/beginner/)
 - [AppCache browser support from caniuse](http://caniuse.com/#search=appcache)
 
 ## Service Worker
 
-This is a part of ES2015 specifications, only some modern browsers support it.
+⚠️ **Service workers only works when using _HTTPS_, for security reasons.**
+_Having modified network requests wide open to man in the middle attacks would
+be really bad_.
+An exception exist for ``http://localhost`` to help you during development.
```
ben-eb: *exists*

docs/content/docs/advanced/offline-browsing.md
```diff
@@ -1,105 +1,200 @@
...
+
+Service Worker are a specific Web worker that allows more flexible behavior
+than Appcache.
+For example, you can easily only cache a single HTML entry point with
```
ben-eb: I would drop the "only" here (as the meaning is conveyed through "single")

docs/content/docs/advanced/offline-browsing.md
```diff
@@ -1,105 +1,200 @@
...
+> - [from MDN](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API/Using_Service_Workers#Registering_your_worker)
+
+### Network first, cache as fallback
+
+⚠️ To always offer up to date and fresh content, we choose to use a
+"network first" approach. Cache will only be used as fallback (eg: no internet
+access).
 
-## How to enable offline support
+## How to enable offline supports for Phenomic?
```
ben-eb: *support* reads better here. 😄

docs/content/docs/advanced/offline-browsing.md
```diff
@@ -1,105 +1,200 @@
...
+### ``"offline": false``
+
+Default value.
+Disable `offline` support.
+
+### ``"offline": true``
 
-- `false`: Default value. Disable `offline` support.
-- `true`: Use default globby patterns `[ "**", "!**/*.html", "index.html" ]`
+Use default configuration (see below) which will cache every resources
```
ben-eb: Replace every with *all*.

thangngoc89:
The code LGTM :+1:

PR #571, opened 2016-07-17
Change import order in LayoutContainer

AdamQuadmon:
The import order affect css loading order.

### DIFF
--- a/boilerplate/web_modules/LayoutContainer/index.js
+++ b/boilerplate/web_modules/LayoutContainer/index.js
@@ -1,6 +1,7 @@
 import React, { Component, PropTypes } from "react"
 import Helmet from "react-helmet"
 
+// Import global CSS before other components and their styles
 import "./index.global.css"
 import styles from "./index.css"
 
--- a/boilerplate/web_modules/LayoutContainer/index.js
+++ b/boilerplate/web_modules/LayoutContainer/index.js
@@ -1,11 +1,11 @@
 import React, { Component, PropTypes } from "react"
 import Helmet from "react-helmet"
 
-import Header from "../Header"
-import Footer from "../Footer"
-
 import "./index.global.css"
 import styles from "./index.css"
+
+import Header from "../Header"
+import Footer from "../Footer"
 
 export default class Layout extends Component {
 


thangngoc89:
Hi, thanks for your PR. Could you give me some details about this change? Why order matter?

AdamQuadmon:
If you look at the output of a page you'll find Header and Footer css styles are loaded before the LayoutContainer ones.

I don't think is a big issue if you use CSS Modules the right way, but can be confusing.

Usually I expect a global normalize.css to be loaded before any other style.
This was the philosophy of Cascading Style Sheet.

thangngoc89:
I understand your concern. But Header and Footer component use CSS Modules. So this is not a necessary change.

> Usually I expect a global normalize.css to be loaded before any other style.

Well. If you work with Webpack long enough. CSS loading order is a mess. You can't guarantee which file will be loaded before which.

If this PR can eliminate confusion for new user. I'm hapy to merge it. 

cc @MoOx

AdamQuadmon:
I'm new to webpack, react, css modules, postcss and almost everything in this stack, and each time I put my hands on it I regret myself to went this way.

I said "this is the future so I need to go this way", but I see confusion and over engineering everywhere.

I needed to change also IDE to work on this stack!

If you say so maybe there is something evil going on here.

I'd like to know and understand each piece of my puzzle, I don't like when (black) magic happen in an unpredictable way.

This is just a little insignificant fix that doesn't fix anything but at least now my reset code appear where I supposed it to be!

thangngoc89:
@AdamQuadmon YESSSS. This is the future. 

FYI: https://github.com/webpack/webpack/issues/215

So when using webpack I recommend that you don't use the "Cascading" of CSS. I have to `!important` everything when I need to overide a global styling.

AdamQuadmon:
I see it.

Things are changing a lot (but I'm not sure for good).
Still feeling lost in the postcss word! Maybe the "post" here means CSS is dead or something like that.
PastCss, DeC(e)ss, RestCss(inPeace)...
Hope to find some guidelines to update my knowledge and habits.

Thanks for your patience with this old coder!

MoOx:
@AdamQuadmon thank you for raising this issue.

CSS loading order is not a mess due to webpack, it's due to how we load the components in the page.

Keep in mind that the same issue might happen with browserify (if you use it to load styles) since you won't really control how components are loaded.

For example, if you open one route that load a component, then another that load some others components (and styles) - if you do the opposite, you will have the CSS behing loaded in the browser with another order - so might affect your global rendering if you rely on global styles - that's a reason why inline styles and CSS modules have been created.

@AdamQuadmon Would you mind adding a simple comment on top of the global.css import like `// Importing global CSS before other components an their styles` ?

Another possibility is to import styles like normalize in the script/phenomic.browser.js using `import "normalize.css"` but you will need to tweaks the webpack loaders config for this. Maybe we should add this ability in the default boilerplate (css from node_modules as global css by default? This might create issue with components from npm that use css modules, but those might require some webpack config adjustements anyway)

AdamQuadmon:
I'm doing:

`LayoutController/index.js`
```js
import "./reset.global.css"
import "./index.global.css"
```
(I split the file to better read the output code)

`reset.global.css`
```
@import "normalize.css";
@import "cssrecipes-defaults";
```

`webpack.config.babel.js`
```
//...
        {
          test: [ /\.css$/, /\.scss$/ ],
          include: path.resolve(__dirname, "node_modules"),
          loader: ExtractTextPlugin.extract(
            "style-loader",
            [ "css-loader", "postcss-loader", "sass-loader" ].join("!"),
          ),
        },
// ...
postcss: () => [
      require("stylelint")({
        ignoreFiles: path.dirname(__dirname, "node_modules/**/*.css"),
      }),
      require("postcss-reporter")(),
      require("postcss-import"),     
    ],
```

I'm trying to leverage on PostCss as much as I can, I was used to stuff like Bootstrap and Foundation, and I'm trying to change my mindset

MoOx:
@AdamQuadmon that's a possibility indeed.

Do you want to handle the small add I requested for the comment?

AdamQuadmon:
@MoOx done

MoOx:
Thanks @AdamQuadmon!

Issue #617, opened 2016-08-15
Add some CSS variables in the default theme


MoOx:
As an example, via http://cssnext.io/usage/#features (simplest way) or via CSS modules `@value` (but I don't like this totally non standard syntax).

Following @revolunet's request on the chat.

revolunet:
How should this be implemented  ? a separate `cssVariables.js` file required by webpack/postcss-cssnext ? 
Also, where should we store shared "mixins" for composes ?

MoOx:
Currently yeah you could pass options here https://github.com/MoOx/phenomic/blob/2b44e19bfa095d2a4bb512c122dbb24380a83ef1/boilerplate/webpack.config.babel.js#L150, imported from another file works.

For mixins, I think if you need more than a simple value, you should probably write a UI component ;)

PR #618, opened 2016-08-15
Boilerplate: add cssnext features example

revolunet:
Following #617

not sure about the file naming and position though


MoOx:
Thanks a lot!
I will wait for another feedback to be sure that's the right approach we want to recommend.
If it's ok for others we can probably merge this :)

revolunet:
there must be a better that could also include mixin from the same file ?

revolunet:
i'll add an entry to the FAQ instead of making it default

PR #622, opened 2016-08-16
Faq: add LESS info

revolunet:
add pointers to have LESS support

### DIFF
--- a/docs/content/docs/faq/less.md
+++ b/docs/content/docs/faq/less.md
@@ -1,0 +1,23 @@
+---
+title: How to add LESS support
+---
+
+So if you want to add some global LESS to your phenomic (for components, you'd better use scoped CSS modules) :
+
+ - Add a `layouts/LayoutContainer/index.global.less` file
+ - Require it from `layouts/LayoutContainer/index.global.less` with `import "./index.global.less"`
+ - Install dependencies : `npm i -S less less-loader`
+
+In `webpack.config.babel.js`, add a LESS module loader :
+
+```js
+  {
+    test: /\.global\.less$/,
+    include: path.resolve(__dirname, "web_modules"),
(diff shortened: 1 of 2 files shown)


docs/content/docs/faq/less.md
```diff
@@ -0,0 +1,23 @@
+---
+title: How to add LESS support
+---
+
+So if you want to add some global LESS to your phenomic (for components, you'd better use scoped CSS modules) :
+
+ - add a `layouts/LayoutContainer/index.global.less` file
```
thangngoc89: Could you use uppercase for the first letter?

thangngoc89:
We can create a new section in the docs and call it receipt. It's good for stuff like this.

And you should add using LESS in local scope.

revolunet:
fixed :)

what do you mean by "using LESS in local scope." ?

thangngoc89:
@revolunet I mean using LESS with css modules. (local-scoped)

revolunet:
ah yes, i'd love a pointer to an example then :)

thangngoc89:
Just duplicate these lines https://github.com/MoOx/phenomic/blob/2b44e19bfa095d2a4bb512c122dbb24380a83ef1/docs/webpack.config.babel.js#L53-L65

Replace postcss-loader with less-loader and modify the regex

revolunet:
thanks :)

MoOx:
I think I am going to explode the FAQ to move response in the appropriate part of the doc. I will still keep the FAQ list, but move response that can be. Will try to handle that.

MoOx:
Going to make some adjustment & move that in the "Getting started" as it's a must have! Thank you! ❤️

MoOx:
Merged by hand as there is conflict ;)

PR #623, opened 2016-08-16
Faq: add CSS Modules shared variables example

revolunet:


thangngoc89:
the code is good but you get the idea wrong. The title should be how to add global variables with cssnext (actually it's postcss-loader). 

CSS Modules is different thing and it's a seperated from cssnext. You can use CSS Modules with any CSS-preprocessor you like,SASS, LESS, Stylus, <insert your css-preprocessor here>. 

You can read more about it here https://github.com/css-modules/css-modules

revolunet:
I fixed the text & hope its better :)

If i have configured LESS support in CSS modules, what should have been the syntax to get the variable from LESS files ? same `var(--name)` ?

thangngoc89:
I dont know LESS. But when using SASS, this is what I would do (straightforward language's syntax) 

```
$some-var = red
```

And in other file, import it with sass syntax.

thangngoc89:
LGTM. cc @MoOx

MoOx:
Going to make some adjustement & move that in the "Getting started" as it's a must have! Thank you! ❤️

PR #633, opened 2016-08-18
0.16

MoOx:


docs/web_modules/Showcase/index.js
```diff
@@ -0,0 +1,162 @@
...
+      return true
+    }
+    phenomicDocs = item
+    return false
+  })
+  if (phenomicDocs) {
+    currentList.push(phenomicDocs)
+  }
+
+  const addYourOwn = (
```
thangngoc89: You should put this component outside. This will run re-define each time the component was rendered

src/_utils/array-unique/index.js
```diff
@@ -0,0 +1,5 @@
+// @ flow
+// flow is not enabled because of the issue with set and spread
+// https://github.com/facebook/flow/issues/1059
+
+export default (array) => [ ...(new Set(array)) ]
```
thangngoc89: :100: 1 liner array unique

src/static/routes-to-urls/__tests__/index.js
```diff
@@ -0,0 +1,104 @@
+// @ flow
```
thangngoc89: Is this a mistake ? This doesn't enable flow

docs/web_modules/Showcase/index.js
MoOx: I will do that next time I will open this file, no big deal :) #lazyMoOx

src/static/routes-to-urls/__tests__/index.js
MoOx: Yeah probably a mistake. No big deal, no tests are "flowed" yet. Will revisit when refactoring for monorepo.

PR #635, opened 2016-08-18
Remove UglifyJSPlugin from node build

thangngoc89:
No real perf gain on but I think this is a good thing to do

### DIFF
--- a/src/builder/webpack/config.node.js
+++ b/src/builder/webpack/config.node.js
@@ -1,16 +1,18 @@
 // @flow
 
 import { join, basename } from "path"
-import { BannerPlugin } from "webpack"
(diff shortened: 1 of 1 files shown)


src/builder/webpack/config.node.js
```diff
@@ -1,7 +1,7 @@
 // @flow
 
 import { join, basename } from "path"
-import { BannerPlugin } from "webpack"
```
MoOx: Why not `import { BannerPlugin, optimize: { UglifyJsPlugin } } from "webpack"` ?

thangngoc89: I did that but eslint throw syntax error

Issue #641, opened 2016-08-22
Build context


DavidWells:
I am running into a problem with a third party module (from auth0), where one of their dependancies is referencing `window` and `navigator` and is causing the phenomic node build to fail. (`npm start` dev works fine, but node static build chokes)

I'm unable to differentiate the node vs browser build from my webpack config in my site. 

It looks like that would have to happen deeper at https://github.com/MoOx/phenomic/blob/master/src/builder/webpack/config.common.js#L25 

I'm proposing adding a `PHENOMIC_BUILD_CONTEXT` variable here so we can use conditionals in our code to include none isomorphic friendly libs.

In my code, I'd like to do this: 
```js
// dynamic require in the phenomic site
if (process.env.PHENOMIC_BUILD_CONTEXT === 'browser') {
  const ModuleReferencingWindow = require('non-isomorphic-compatible-lib')
} else {
  // node context, don't require the code for isomorphic build
  const ModuleReferencingWindow = {}
}
```

What do you guys think?

I'm working on adding Auth to our new site and this is a core piece of it =). Going to discuss this in the phenomic demo I'm doing at the react meetup later this month

thangngoc89:
my way of checking for browser/node: 

if (typeof window !== 'undefined')

Could you try that first? Afaik, that's only way for isomorphic app. I don't want to introduce a different way of doing this in phenomic

DavidWells:
I tried that. It didn't work. Node kept choking on the node dependancy (out of my control)

I will try again.

But the webpack provide in browser.webpack fixed the issue for me

DavidWells:
Lolz. Nevermind. @thangngoc89 you are right =P

I was wrapping my module incorrectly

PR #642, opened 2016-08-22
Add browser context to browser build

DavidWells:
#641

Add browser context global for webpack to not choke on node static build

Example usage here: https://gist.github.com/DavidWells/94df55e693d8ecf6de22576cf397f6e2

This fixes older npm packages referencing `window`, `document` etc. 

```
if (process.env.PHENOMIC_BROWSER_CONTEXT) {
  // is browser, include package that breaks in node
  var Auth0Lock = require('auth0-lock').default
}
```

### DIFF
--- a/src/builder/webpack/config.browser.js
+++ b/src/builder/webpack/config.browser.js
@@ -1,12 +1,12 @@
 // @flow
 
 import { join } from "path"
-
+import { DefinePlugin } from "webpack"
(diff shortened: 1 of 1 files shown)


DavidWells:
typeof window !== 'undefined' fixed this... I'm an idiot lolz

MoOx:
I might be open to this if someone else report a similar "issue" (ie: don't have to idea to test window type).

PR #690, opened 2016-08-30
Add className support to page container

DavidWells:
I need access to the page container to apply a class.

My use case is to apply some fancy min height css to the page container to make it fit in viewport always

```
.pageContainer {
 min-height:  min-height: calc(100vh - $navHeight - $footerHeight);
}
```

image

### DIFF
--- a/src/PageContainer/component.js
+++ b/src/PageContainer/component.js
@@ -15,6 +15,7 @@ type Props = {
     splat: string,
   },
   layouts: Object,
+  className: string,
   defaultLayout: string,
   getPage: Function,
   setPageNotFound: Function,
@@ -234,7 +235,7 @@ class PageContainer extends Component<DefaultProps, Props, void> {
     const Layout = getLayout(page.type, props) || LayoutFallback
 
     return (
-      <div>
+      <div className={props.className}>
         {
           !page.error && page.loading && PageLoading &&
           <PageLoading />


DavidWells:
Hmmm I guess I can wrap my code: https://github.com/serverless/site/blob/master/src/app.js#L48 but I'd like to apply classes to phenomic div too https://github.com/serverless/site/blob/master/src/layouts.js#L15 =)

thangngoc89:
@DavidWells try flexbox. it would be a lot easier 

https://philipwalton.github.io/solved-by-flexbox/demos/sticky-footer/

src/PageContainer/component.js
```diff
@@ -15,6 +15,7 @@ type Props = {
     splat: string,
   },
   layouts: Object,
+  className: string,
```
thangngoc89: We don't want this a requirement. Try `className?: string`

MoOx:
@DavidWells I am not against this change, but indeed, flexbox is the best solution here. You can see it in action here https://phenomic.io/docs/
I am also going to update the default boilerplate (wip) to have this by default (body take full height - header - footer).

thangngoc89:
IMHO, we should allow user to pass their props to PageContainer. How about `{...remainingProps}`. This way, we don't have to merge PR for each specific case like this

MoOx:
Yeah good idea!

DavidWells:
@MoOx not sure if people will want that default flex or not. Aka not sure of side effects it might cause for people who want ugly sites =PP

I'd just spread the props into <PageContainer/>

PR #691, opened 2016-08-31
Removed: ``PageContainer`` does not wrap its child into a ``<div>``

MoOx:
Closes #690 

poke @DavidWells

### DIFF
--- a/src/PageContainer/component.js
+++ b/src/PageContainer/component.js
@@ -233,29 +233,27 @@ class PageContainer extends Component<DefaultProps, Props, void> {
     const LayoutFallback = getLayout(props.defaultLayout, props)
     const Layout = getLayout(page.type, props) || LayoutFallback
 
-    return (
-      <div>
-        {
-          !page.error && page.loading && PageLoading &&
-          <PageLoading />
-        }
-        {
-          !!page.error && !PageError &&
+    if (page.error) {
+      if (!PageError) {
+        return (
           <div style={ { "text-align": "center" } }>
             <h1>{ page.error }</h1>
             <p>{ page.errorText }</p>
           </div>
-        }
-        {
-          !!page.error && PageError &&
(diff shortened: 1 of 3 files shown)


MoOx:
I think I have a better idea: we should remove this div. It's useless. We should not render multiple components in PageContainer anyway.
Does this looks good to you?

thangngoc89:
LGTM

MoOx:
Updated. What do you think?

src/PageContainer/component.js
```diff
@@ -233,29 +233,27 @@ class PageContainer extends Component<DefaultProps, Props, void> {
...
+    else {
+      if (page.loading && PageLoading) {
+        return <PageLoading />
+      }
+      else if (Layout) {
+        return <Layout ref={ this.saveContentRef } { ...page } />
+      }
+    }
+
+    return null
```
thangngoc89: Is this valid ? Never return null from React render

thangngoc89:
It's much cleaner now. But I saw these changes in the [feat.prepare-theme-base](https://github.com/MoOx/phenomic/tree/feat.prepare-theme-base) too. Rebased?

src/PageContainer/component.js
MoOx: Yeah it's totally valid, React render null as... nothing! Handy for component that can skip their render themselves.
But in our case in theory, this should not happen :)

MoOx:
Last commit of this branch was a totally mess, it was to show this to a friend. I am editing this branch ;)

src/PageContainer/component.js
thangngoc89: I learned new thing today. Thanks :100:

MoOx: This can be convenient for stuff like a `<LoadingIndicator loading={ boolean } />` :D

PR #696, opened 2016-08-31
Only clear /dist on prod build

DavidWells:
This is running everytime and sometimes I want to look at the /dist without having to rebuild =P

`phenomic start` clears the dist folder and doesn't need to.

Only the `phenomic build` should clear the dist.

### DIFF
--- a/src/builder/index.js
+++ b/src/builder/index.js
@@ -35,7 +35,12 @@ export default function(config: Object): void {
   config.webpackConfigNode = webpackConfigNode(config)
 
   const destination = join(config.cwd, config.destination)
-  fs.emptyDirSync(destination)
+  
+  if (config.production) {
+    // If production build, clear /dist folder
+    fs.emptyDirSync(destination)
+  }
+
 
   if (config.static) {
     // Copy static assets to build folder


MoOx:
/dist is not used in dev, so that can only create confusion to keep an out of date /dist folder during dev mode. Why do you want to keep /dist ?

DavidWells:
In my dev workflow I typically only clear out `dist` or `build` folders
when I am building for production.

I sometimes want to reference/see the `/dist` files contents and have to
rebuild to see stuff.

It's also a pain when I'm showing more and more people phenomic and I try
to explain how the `/content` maps to `/dist` and demo hot-reloading etc
but that wipes out the `/dist` and I have to stop and rebuild and wait.

I want to be able to demo phenomic's local dev setup aka `phenomic start`
and show them the dist folder in 2 seconds instead of waiting for that
build =P

DavidWells:
I don't think the outdated `dist` folder hurts anything by hanging around in local dev but maybe that's just my opinion. You should know you need to `npm run build` to deploy new changes.

Anywho, this is a convenience thing more than anything

DavidWells:
On second thought, this seemed to be causing issues for me.

When `/dist` wasn't being removed. Local dev wasn't hot-reloading for some reason.  (not sure why)

Closing this.

MoOx:
I just left my computer. Will review and merge tomorrow then :)

MoOx:
The problem here, it that webpack-dev-server can serve dist before builded assets stored in memory...

PR #698, opened 2016-08-31
Update boilerplate to be a base theme

### LINKED ISSUE #433: Normalize "phenomic(/lib/*)" public API
Currently some stuff are in `"phenomic/lib/*"` and since #432 we introduced an utility in "phenomic" itself.
When we will rely on webpack 2, it will be safe to make move lib/* interfaces as exported value in "phenomic".
We will have one exception to handle for the loader name that must be a string to a pass, but we can export the string too (eg: `import { contentLoader } from "phenomic"` and later use a template string in webpack config).
We will have to warn and deprecate all lib/* interfaces (move lib to another place, create a lib that require new files + warn when imported).

### LINKED ISSUE #444: Boilerplate: Rename LayoutContainer to AppContainer
Following https://twitter.com/MoOx/status/728210032093990912

### LINKED ISSUE #529: Change the default structure of the project
`web_modules` might be confusing.
Something like

```
/
  /app?
  /layouts
  /modules? /components?
  [/themes?](#508)
```

Need to think more tho.

MoOx:
Closes #529 
Closes #444 
Will closes #433


thangngoc89:
So far so good, I didn't see any different from this and the old boilerplate. Am I right ? Just out of curiosity, do you have plan to let user swap out **compatible** theme with 1 line of config ?

MoOx:
> So far so good, I didn't see any different from this and the old boilerplate.

You probably didn't look enough. There are a lot of noticeable differences if you look at the tree.

> Am I right ? Just out of curiosity, do you have plan to let user swap out compatible theme with 1 line of config ?

On the long run yes, but for now we are far from that (maybe not that far but still).

MoOx:
PR updated. Should be good now.

themes/phenomic-theme-base/src/components/Container/index.js
```diff
@@ -0,0 +1,18 @@
...
+
+export default class Container extends Component {
+
+  static propTypes = {
+    children: PropTypes.oneOfType([ PropTypes.array, PropTypes.object ]),
+  };
+
+  render() {
+    return (
+      <div className={ styles.container }>
```
thangngoc89: I think this is unnecessary, can we just move the styling part to AppContainer ?

themes/phenomic-theme-base/src/components/DefaultHeadMeta/index.js
```diff
@@ -23,10 +12,12 @@ export default class Layout extends Component {
       pkg,
     } = this.context.metadata
 
+    /* eslint-disable react/jsx-key */
     return (
-      <div className={ styles.layout }>
+      <div hidden>
```
thangngoc89: Haha. This will upset IE. We should add a notice about html5 shim if user want to support IE < 9

thangngoc89:
Except for my comments. LGTM

themes/phenomic-theme-base/src/components/Container/index.js
MoOx: This is to keep a simple AppContainer without any CSS.

themes/phenomic-theme-base/src/components/DefaultHeadMeta/index.js
MoOx: no big deal, the div is empty so should not render anything anyway.

themes/phenomic-theme-base/src/components/Container/index.js
thangngoc89: OK. :+1:

PR #721, opened 2016-09-04
Define index route in frontmatter

### LINKED ISSUE #673: Define root index of the project via `route` parameter
Looks like there is no way you can define a route like this 

```
---
route: /
---
```

and expect it to becomes the root of your project.

I tried with several alternative route like `/index.html`, `index.html` but no luck. 

Investigating this

thangngoc89:
Fix #673

### DIFF
--- a/src/phenomic-loader/__tests__/index.js
+++ b/src/phenomic-loader/__tests__/index.js
@@ -86,6 +86,17 @@ test.cb("phenomic loader", (t) => {
         "should create a proper json for custom route with an extension"
       )
 
+      const customRouteRootIndex = stats.compilation.assets[
+        //    fixtures/custom-route-root-index.md
+        // -> fixtures/index.html
+        "index.html" +
+        ".8817d2a1fab9dfb9b4b52cd6ee7529ab.json"
+      ]
+      t.truthy(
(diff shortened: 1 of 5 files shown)


src/phenomic-loader/__tests__/index.js
```diff
@@ -86,6 +86,17 @@ test.cb("phenomic loader", (t) => {
...
+      const customRouteRootIndex = stats.compilation.assets[
+        //    fixtures/custom-route-root-index.md
+        // -> fixtures/index.html
+        "index.html" +
+        ".8817d2a1fab9dfb9b4b52cd6ee7529ab.json"
+      ]
+      t.truthy(
+        customRouteRootIndex && customRouteRootIndex._value,
+        "should create a proper json for custom route with an extension"
+      )
```
MoOx: This test does not seems to add some value here. What is this supposed to check?

thangngoc89: Before this PR, webpack doesn't emit `/index.html` file and its json file.

MoOx: haha OUCH.

PR #723, opened 2016-09-05
#679, replace default message '...'

### LINKED ISSUE #679: FAQ: why did I get a page with "..." when running "npm start"?

kevin940726:
Fix #679 .


thangngoc89:
Thanks a lot. LGTM

MoOx:
What is the benefit of this? At least we should show a spinner or something. Maybe a "splash screen". Showing and empty page looks even worst to me. Bad UX in case of a JS fatal error...

PR #726, opened 2016-09-05
Fix broken link

kevin940726:
Fix a broken link. Also, `CNAME` will only be needed if one want to set the sub-domain stuff, am i right? This line seems a little confusing that it it seems that we **should** set `CNAME` to true.

### DIFF
--- a/docs/content/docs/usage/gh-pages.md
+++ b/docs/content/docs/usage/gh-pages.md
@@ -3,7 +3,7 @@ title: Automatic deployment on GitHub Pages
 ---
 
 **Be sure that you use the ``phenomic.CNAME`` option in your
-[configuration](./configuration/)**
+[configuration](../configuration/)**
 
 You will have multiple possibilities to deploy your `dist` folder on the
 `gh-pages` branch.


MoOx:
What do you call "the sub-domain stuff" ?

kevin940726:
I meant [this](https://help.github.com/articles/setting-up-a-custom-subdomain/) 😅 .

MoOx:
Not only. Using the CNAME option is to generate a CNAME file according to your homepage url. So this work for all domains other than *.github.io.

MoOx:
Thank you for this fix :)

kevin940726:
Custom domain should be more precisely 😅 ?

So how do you think of the documentation part? Do we need to change the wording to make it clear?

MoOx:
Not sure what do you want to change. The current CNAME option is "just" to generate a CNAME file from your "package.json/homepage" field. 
But feel free to propose some enhancement, help is always appreciated! ❤️

kevin940726:
I actually meant this line:
> Be sure that you use the `phenomic.CNAME` option in your configuration

Maybe it's just me but the word `use` here seems a little confusing, don't you think? At first glance I would just go and set it to true without thinking 😅 .
Maybe something like...
> If you want to deploy to a custom domain, be sure to use the `phenomic.CNAME` option in your configuration

MoOx:
You are totally right. Can you make a PR for this? Thanks in advance!

PR #728, opened 2016-09-05
Fix wording to make it clear for CNAME option when deploying.

kevin940726:
See #726 .


MoOx:
Thanks! 🎉

MoOx:
(website will be updated with the next release)

Issue #663, opened 2016-08-24
Docs: How do I add a featured image in the markdown file?


MoOx:

drFabio:
What is this about ?
I am already adding an image by adding on the metadata and extending PagePreview for example to display it on the list. Is something like this or something else?

MoOx:
Something like that yes. Use a field in the metadata and then just use this field in a component :)

MoOx:
Closing in favor of #736. Thanks @drFabio!

PR #736, opened 2016-09-08
663 Add feature image on markdown

drFabio:
Faq entry about markdwon feature images , related to [663](https://github.com/MoOx/phenomic/issues/663).
Not sure about the title of the link and the post OK abou the content but inputs are welcome.

### DIFF
--- a/docs/content/docs/faq/images.md
+++ b/docs/content/docs/faq/images.md
@@ -1,0 +1,48 @@
+---
+title: Phenomic FAQ about Images 
+---
+Besides being able to add the image on the markdown itself it's possible to add it on the post metadata and use it later.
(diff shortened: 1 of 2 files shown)


MoOx:
Thanks a lot for your contribution! That's a very nice idea to add stuff like that <3

Like explained on the chat, we are currently thinking on a better approach to sort all FAQ entries, maybe in subsection with dedicated pages. Your content may even end up in the Getting Started page!

So this will definitely be merged soon, can't tell when, but it will be ;)

MoOx:
Going to make some adjustement & move that in the "Getting started" as it's a nice example! Thank you! ❤️

Issue #598, opened 2016-08-04
Explode as a monorepo


MoOx:
This might require a lot of work but maybe splitting into something like this

phenomic-core
phenomic-content-loader
phenomic-content-loader-plugin-*
phenomic-content-loader-preset-\* (bundle of plugins, for example phenomic-content-loader-preset-markdown)
phenomic-static-rendering

Just an idea :)

thangngoc89:
What is `core` and `static-rendering` ? Still, this is a pretty good ideas and a good chance to clean up and organize the codebase. I hate it when the `src` folder has too many subfolders

MoOx:
core will be, well the core part (not sure what this will be yet) and static-rendering will be code for the static html generation part. But maybe this should be in the core as well.

spoiler: with monorepo, we might have even more (sub)folders ;)

MoOx:
To check: 
- http://lernajs.io/ (I saw so bad feedback: slow https://github.com/pouchdb/pouchdb/issues/5545)
- https://formidable.com/open-source/builder/ (maybe a bit weird)
- https://github.com/northbrookjs/northbrook (inspired by cyclejs monorepo)
- some infos
  - https://github.com/pouchdb/pouchdb/pull/5178
  - https://github.com/boennemann/alle#readme

thangngoc89:
Biggest trade: no greenkeeper (We are suffering this already with docs and boilerplate)

About the structure, I think we should keep the core of Phenomic in one piece at first. only break plugins section.

MoOx:
There is solution for greekeeper + greenkeeper plans to handle monorepo as well.

MoOx:
Might be interesting for consuming git repo https://github.com/facebook/jest/blob/2fcae4df8256ed00e3b995cf7ea2957b8766e7a5/scripts/postinstall.js

MoOx:
@thangngoc89 shared this https://github.com/knitjs/knit looks promising

bloodyowl:
Going to be fixed with #925

PR #737, opened 2016-09-08
Switch to Webpack 2

MoOx:
~~We are blocked by https://github.com/mzgoddard/hard-source-webpack-plugin/issues/27 (hmr + cache are not friends for now).~~

~~Otherwise~~ all green!


themes/phenomic-theme-base/scripts/phenomic.browser.js
```diff
@@ -1,26 +1,32 @@
+// Hot loading HRM Patch
+import "react-hot-loader/patch"
```
thangngoc89: Can UglifyJS remove this in production mode? Why don't you add this to devEntries in dev-server. We shouldn't require any changes in the boilerplate from user.

MoOx: Currently, user need to explicitely use module.hot.accept (see below), so just moving this call is useless. Also if people have issue with RHL@3, they can replace it (or remove it). I will make some tests later to provides better default, but this will bring this kind of module as a dep (and we will need to use an option to enable/disable this kind of stuff).

And don't worry, the patch already have a dev/prod hook itself ;)

MoOx: Previously, the call was still explicit in babel conf (in the webpack config)

thangngoc89: Ok. Got it.

thangngoc89:
Hard source webpack plugin has a fix for us

MoOx:
@thangngoc89 I just installed 0.0.35 but it does not fix things for us :(

thangngoc89:
Ouch.

MoOx:
Wait a sec. It works with cache on (first pass, without no cache yet), but does not work for the 2nd build (when cache is used).

thangngoc89:
Could you report this back to HardSource?

MoOx:
Done in the PR.

MoOx:
Also it seems that webpack 2 does not magically avoid having unwanted stuff like everything related to the rendering engine (lowlight, highlight, lodash etc) which sucks. Will have to investigate.

thangngoc89:
> Also it seems that webpack 2 does not magically avoid having unwanted stuff like everything related to the rendering engine (lowlight, highlight, lodash etc) which sucks. Will have to investigate. 

It's because you're using transpiled code when importing `phenomic`.

MoOx:
You are right. Then I will have to separate "phenomic" package before handling #598.

MoOx:
- "phenomic" import now do not have loader plugins & presets anymore
- hmr seems to work

MoOx:
Problem: Hot loading seems slower with webpack 2 + cache.
And also break our collection runtime "cache"... :'(

HMR + no cache

screen shot 2016-09-11 at 00 11 02

HMR + new cache

screen shot 2016-09-11 at 00 12 07

HMR + cache

1.1s the first time, but no collection is generated :@
This break our project since it seems that this file is cached https://github.com/MoOx/phenomic/blob/08bd04feb2254891caf809e383fca0f6608670c1/src/loader/cache.js and it should not because we use this as a "runtime" cache...

Poke @mzgoddard if you have an idea about this performance issue & this cache issue...

mzgoddard:
@MoOx I have a pretty good guess what is adding time after HardSource invalidates the cache. After the first build after that message from HardSource, if you restarted your server the build times will go down. Basically HardSource is serializing **all** of your modules every time after that first build, as opposed to just the ones that changed, because it doesn't know they're already serialized. I'm gonna make HardSource just a little smarter and that should keep it from re-serializing everything.

MoOx:
@mzgoddard the thing is, during the start (with existing cache), a file is aggressively cached https://github.com/MoOx/phenomic/blob/08bd04feb2254891caf809e383fca0f6608670c1/src/loader/cache.js and our loader have no effect on it (it's currently adding all pages data into this object, but in practise, nothing is added, so the website have no page data :'(.

Any idea how we can fix this issue?

MoOx:
Perf seems fixed with latest version of hard-source-webpack-plugin@0.0.37, but our runtime cache :/ Not sure how we can handle that... Or maybe it has something to do with require.context() caching... Will try to find where the issue comes from tonight.

mzgoddard:
@MoOx I think you need to move the cache and feed creation into a plugin. You want part of how ExtractTextWebpackPlugin works. The stuff you are storing on cache should be stored on the module in the meta attribute that HardSource serializes. Then with that stored in meta, whether its the first run with HardSource or a different run from a preexisting cache your plugin can build your feeds during a compilation step like `additional-assets`. This should also remove what looks like race conditions to me with debouncing the feed work. (Is there an existing plugin that handles feed assets being in multiple built modules if the timeout debouncer runs with more upcoming modules that will run it a second time, or between multiple builds where a different module will have the feeds built the second time?)

So the first part of this is storing the info for your feeds [like ExtractText](https://github.com/webpack/extract-text-webpack-plugin/blob/master/index.js#L212-L222):

```js
var fs = require('fs');

// Use the path of the plugin/loader directory to avoid conflicts on the loaderContext.
var NS = fs.realpathSync(__dirname);

module.exports = function PhenomicLoaderPlugin(options) {
  this.options = options;
};

PhenomicLoaderPlugin.prototype.apply = function(compiler) {
  var options = this.options;

  compiler.plugin('compilation', function(compilation, params) {
    compilation.plugin('normal-module-loader', function(loaderContext, module) {
      loaderContext[NS] = function(loaderResult) {
        module.meta[NS] = loaderResult;
      };
    });

    compilation.plugin('additional-assets', function(callback) {
      // Collect results
      var results = compilation.modules.map(function(module) {
        return module.meta[NS];
      });

      // Build feeds
      Object.keys(options.feeds).forEach(function(name) {
        compilation.assets[name] = feed({
          // ...
        });
      });

      // Make sure to call callback or webpack will just stop because Node has
      // nothing left in the event queue to do.
      callback();
    });
  });
};
```

Then in place of storing on the cache you need to call that `NS` method in your loader.

```js
var fs = require('fs');

// Use the path of the plugin/loader directory to avoid conflicts on the loaderContext.
var NS = fs.realpathSync(__dirname);

module.exports = function() {
  // Do needed loader work
  // ...

  // Make sure this data is ok to be JSON stringified and parsed.
  this[NS](result);

  // Any further loader work
  // ...
};
```

MoOx:
The thing is: our loader is not just doing a feed, it's also generating json for each entry (md file) an return an object with the path to the emited file. And this cache is used for the static build, which avoid us to read data twice (but maybe we should read it twice...)
I will try to investigate but I guess we might end with a refactoring somewhere so your plugin work correctly with our project :(

mzgoddard:
@MoOx keep using your loader to do those parts, to generate json per file and return the path to that file. Its just the feeds that need to be made by a plugin. [This part of the loader](https://github.com/MoOx/phenomic/blob/08bd04feb2254891caf809e383fca0f6608670c1/src/loader/index.js#L67-L109) needs to be done by a plugin. You don't need to read data twice, the plugin exposes a function your loader uses to save that data so the data can be reused by the plugin later. You could have a function on your plugin to expose the saved info to any other part of your build.

MoOx:
@mzgoddard your advices and webpack expertise were really precious. Now everything works as expected. Thank you so much for all your contributions.

Issue #647, opened 2016-08-22
Ignore file-ignored warnings from stylelint


kennu:
I'm trying to ignore Bootstrap CSS (in node_modules) by adding

    "stylelint": {
      "ignoreFiles": "**/bootstrap/**"
    }

It works, but it causes the Phenomic build process to bark out a 500 line warning message every time I edit something:

    WARNING in ./~/css-loader!./~/postcss-loader!./~/bootstrap/dist/css/bootstrap.css
    stylelint: This file is ignored

Couldn't figure out a way to ignore CSS files without the warning.

MoOx:
Ouch. I have to admit this sucks. Need to check who is the culprit.
That can be stylelint itself or webpack... poke @davidtheclark @jeddy3

MoOx:
@kennu Maybe you can remove stylelint for this loader section? Should be straight forward.
Not sure if stylelint maintainers want to remove this warning :/

kennu:
@MoOx Removed it, thanks for the advice. I'm still learning the framework so it was difficult to know where everything is enabled/disabled. But it's a really nice framework. :-)

davidtheclark:
As of v7 stylelint reporter only tells you about ignored files if you use the verbose option. https://github.com/stylelint/stylelint/blob/master/CHANGELOG.md#700 So this is either due to an outdated version or the webpack loader doing its own thing, I think.

MoOx:
Thanks a lot for the explanation. We currently use an outdated version (6.8.x). Sorry!
I will close this when we will udpate to 7.x.

MoOx:
Closed by #741

Issue #718, opened 2016-09-03
Update eslint(-*) & stylelint(-*) to latest


MoOx:
Cf greenkeeper PRs.

thangngoc89:
Resolved in #741

MoOx:
eslint has not been updated.

thangngoc89:
oops. Sorry

MoOx:
Thanks @comoss!

PR #741, opened 2016-09-12
Update stylelint & stylelint-config-standard. (#718)

ben-eb:
I've stuck with the standard configuration instead of adapting the configuration to fit the CSS, hope this is OK. 😄

### DIFF
--- a/docs/src/components/AppCacheBanner/index.css
+++ b/docs/src/components/AppCacheBanner/index.css
@@ -1,6 +1,6 @@
 .banner {
   position: fixed;
-  position: sticky;
+  position: sticky; /* stylelint-disable-line */
   top: 0;
   left: 0;
   right: 0;
@@ -18,7 +18,6 @@
   border-radius: 100%;
   animation: PhenomicDocsAppCacheBannerRotation 0.6s infinite linear 0.25s;
   opacity: 0;
-
   margin-right: 0.6rem;
 }
 
@@ -27,6 +26,7 @@
     opacity: 1;
     transform: rotate(0deg);
   }
+
   to {
     opacity: 1;
     transform: rotate(359deg);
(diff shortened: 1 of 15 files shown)


thangngoc89:
@ben-eb we follow standard rule. LGTM

MoOx:
The reference should be edited in all places:
- package.json devDeps
- package.json optionalPeerDeps
- docs/package.json

docs/src/components/AppCacheBanner/index.css
```diff
@@ -1,6 +1,6 @@
 .banner {
   position: fixed;
-  position: sticky;
+  position: sticky; /* stylelint-disable-line */
```
MoOx: what is the problem here?

ben-eb: It's a duplicate declaration, Stylelint would prefer that either `fixed` or `sticky` was specified, not both.

MoOx: Hummmm ok... Not sure if that make total sense to get this in the "standard" config since this kind of tricks are handy when you work with CSS but I guess it's not a big deal for us... poke @davidtheclark @jeddy3

ben-eb: > I've stuck with the standard configuration instead of adapting the configuration to fit the CSS

;)

thangngoc89:
This should close #647 as well

docs/src/components/AppCacheBanner/index.css
jeddy3: Hello. I hadn't considered that use case when enabling the `declaration-block-no-duplicate-properties` rule. I had assumed all fallbacks would be handled by tooling.

There's an appropriate option that to make the rule more permissive for this situation: [`ignore: ["consecutive-duplicates-with-different-values"]`](https://github.com/stylelint/stylelint/tree/master/src/rules/declaration-block-no-duplicate-properties#ignore-consecutive-duplicates-with-different-values).

I've created [an issue](https://github.com/stylelint/stylelint-config-standard/issues/42) to discuss using that option by default in the standard config.

MoOx: @jeddy3 thanks for your input!

PR #744, opened 2016-09-14
Use Jest as test runner

thangngoc89:
WARNING: Huge diffs


thangngoc89:
Well, there are 2 integration tests in `integration-tests/cli.js` that need to be fixed. And I have no idea how to fix them. I'll try to spawn those process instead of execute them

thangngoc89:
Help wanted: 

So I have a test file that execute [phenomic cli here](https://github.com/MoOx/phenomic/blob/01495f5f8042e5914e046469c363497bd82c8d70/integration-tests/cli.js#L15)

That script has `require("babel-register")` inside it

But when I run the test, it throws error about Syntax Error on spread operator

https://travis-ci.org/MoOx/phenomic/jobs/159970393#L744-L751

I can run that external script manually without a problem

So my question is: Does Jest interfere with babel-register or something ?

To reproduce the error message:

```
git clone git@github.com:MoOx/phenomic.git
cd phenomic
git checkout jest
npm install
npm run test-phenomic-theme-base
```

thangngoc89:
@MoOx turn out, the whole problem is because Jest set `process.env.NODE_ENV === "test"` and we have a proper babel setting for `test` env. 

I see `development` and `production` is duplicated [here](https://github.com/MoOx/phenomic/blob/b1bbecbfd5d0fe973933e05a44a113771eaea5a6/docs/package.json#L74-L88)

Can I move them to default babel config?

MoOx:
No you cannot. Just duplicate again to "test". if you move to global conf, this will make duplicate for webpack-*. I have written a blog post about this mess, will publish this soon ;)

thangngoc89:
Hum. If that is the case, it's really a mess. Babel should have an option that allow us to choose when and when not to extend babel config. 

FYI, I'll set NODE_ENV to development than ;)

MoOx:
You can duplicate, it's just for us, this won't affect docs & boilerplate setup. ANyway it's just a 3 lines section.

thangngoc89:
Yeah. I see. Now that problem was fixed, we have a bigger problem. 

```
Error: listen EADDRINUSE :::8081
```

Bed time, will look in the in tmr.

thangngoc89:
All's green. Ready for review.

scripts/jest-framework-setup.js
```diff
@@ -0,0 +1,2 @@
+// eslint-disable-next-line
+jasmine.DEFAULT_TIMEOUT_INTERVAL = 30000 // 30s
```
MoOx: what is this?! Does jest use jasmine?!

src/static/to-html/__tests__/index.js
```diff
@@ -31,8 +31,6 @@ test("don't break if there is nothing to transform", async (t) => {
 })
 
 test("writeAllHTMLFiles", (t) => {
-  t.plan(3)
-
```
MoOx: You cannot just remove that. We need to be sure all assertions are called.

src/static/routes-to-urls/__tests__/index.js
```diff
@@ -81,24 +79,24 @@ test("routes to urls", (t) => {
...
-    log: (message) => {
-      t.truthy(message.includes(
-        "It looks like some parameters can't be mapped to create routes:  :lol"
-      ))
-    },
+    const log = jest.fn()
+    routesToUrls(routesNoMatches, collection, { log })
+
+    expect(log).toBeCalled()
+    expect(log.mock.calls[0][0]).toContain(
```
MoOx: That's why I don't like expect. This make us too relying on the test framework :'(

src/redux/modules/__tests__/pages.js
```diff
@@ -25,7 +25,7 @@ test("should have action to handle get", async (t) => {
 })
 
 test("should have action to handle refresh", async (t) => {
-  t.plan(3)
+  // t.plan(3)
```
MoOx: Not an an option. We need to handle that somehow.

src/redux/modules/__tests__/pages.js
```diff
@@ -1,12 +1,12 @@
...
+import test from "jest-ava-api"
 
 import * as module from "../pages"
 import reducer from "../pages"
 
 process.env.PHENOMIC_USER_PATHNAME = "/"
 
 test("should have action to handle get", async (t) => {
-  t.plan(3)
+  // t.plan(3)
```
MoOx: Not an an option. We need to handle that somehow.

src/loader/__tests__/index.js
```diff
@@ -1,169 +1,174 @@
...
+        Object.keys(stats.compilation.assets)
+        .filter((key) => key.endsWith(".json"))
+        .forEach((key) => {
+          const result = JSON.parse(stats.compilation.assets[key]._value)
+          if (result.test) {
+            expect(result.test).toBe("dumb")
+          }
+        })
+        // t.plan(2+5) // 2, err, warn, 5   => array
+        // t.end()
```
MoOx: How is this check made?

package.json
```diff
@@ -189,24 +186,13 @@
     "prephenomic-theme-base-start": "npm -s run transpile && babel-node scripts/phenomic-theme-base.js",
     "phenomic-theme-base-start": "cd themes/phenomic-theme-base && npm start",
     "preall-tests": "npm run lint",
-    "all-tests": "cross-env TESTING=1 npm run tests && npm run test-phenomic-theme-base && npm run docs",
-    "test": "npm run all-tests",
-    "test-with-coverage": "nyc --check-coverage --lines 70 npm run all-tests",
-    "coverage": "nyc report --reporter=text-lcov | coveralls",
+    "all-tests": "cross-env TESTING=1 jest --coverage",
+    "postall-tests": "npm run test-phenomic-theme-base && npm run docs",
+    "test": "cross-env TESTING=1 jest",
```
thangngoc89: Agreed

package.json
```diff
@@ -189,24 +186,13 @@
     "prephenomic-theme-base-start": "npm -s run transpile && babel-node scripts/phenomic-theme-base.js",
     "phenomic-theme-base-start": "cd themes/phenomic-theme-base && npm start",
     "preall-tests": "npm run lint",
-    "all-tests": "cross-env TESTING=1 npm run tests && npm run test-phenomic-theme-base && npm run docs",
-    "test": "npm run all-tests",
-    "test-with-coverage": "nyc --check-coverage --lines 70 npm run all-tests",
-    "coverage": "nyc report --reporter=text-lcov | coveralls",
+    "all-tests": "cross-env TESTING=1 jest --coverage",
+    "postall-tests": "npm run test-phenomic-theme-base && npm run docs",
```
thangngoc89: Will move them back. I move it there in debugging process.

package.json
```diff
@@ -175,9 +174,7 @@
     "lint:js:flow": "flow check",
     "lint:js": "npm-run-all --parallel lint:js:*",
     "lint": "npm-run-all --parallel lint:*",
-    "tests": "ava",
-    "#integration-tests": "needs this order (index test the build, and cli is cleaning it)",
-    "integration-tests": "ava __tests__/index.js && ava __tests__/cli.js",
+    "integration-tests": "jest --config '{\"testRegex\": \"integration-tests\", \"setupTestFrameworkScriptFile\": \"<rootDir>/scripts/jest-framework-setup.js\"}'",
```
thangngoc89: Just realize that jest accept a file path for configuration. Will fix this.

integration-tests/build-result.js
```diff
@@ -1,11 +1,11 @@
-import test from "ava"
```
thangngoc89: This is me trying to make jest run test from this folder only. See https://github.com/facebook/jest/issues/1696

I'll rename it back

src/loader/__tests__/index.js
thangngoc89: Yup. That's why I commented out instead of deleted it

src/redux/modules/__tests__/pages.js
thangngoc89: Yup. That's why I commented out instead of deleted it

src/redux/modules/__tests__/pages.js
thangngoc89: Yup. That's why I commented out instead of deleted it

src/static/routes-to-urls/__tests__/index.js
thangngoc89: We can pull-in expect from npm or chai if you prefer. Jest works great with them. But I am not sure about pretty diff.

src/redux/modules/__tests__/pages.js
thangngoc89: There is no reference in Jest's docs for this kind of validation.

src/static/routes-to-urls/__tests__/index.js
MoOx: I would prefer to get a similar api as what we have currently. jest-ava-api might be still used and improved :)

scripts/jest-framework-setup.js
thangngoc89: Yes. Jest uses jasmine2 underhood.

src/static/routes-to-urls/__tests__/index.js
thangngoc89: Hum... What are you suggesting? AVA doesn't have this built-in. We used sinon for this job.

src/bin/commands/setup/__tests__/inquirer.js
```diff
@@ -1,29 +1,32 @@
...
-    if (code !== 0) {
-      console.error("Script exited with code: " + code)
-      t.fail()
-      t.end()
-    }
-    else {
-      t.pass()
-      t.end()
-    }
+describe("bin > commands > setup > inquirer", () => {
```
MoOx: are describe mandatory? What are they for?

src/static/routes-to-urls/__tests__/index.js
MoOx: Here for example, we were passing a callback directly. No need to use jest mock since you can pass a log function. I am just not sure how to handle the t.plan pattern.

src/bin/commands/setup/__tests__/inquirer.js
thangngoc89: They are not mandatory. They just there to help us know where are those test result comes from

src/static/routes-to-urls/__tests__/index.js
thangngoc89: Yup. We need t.plan or similar to do callback testing like this

package.json
thangngoc89: There is 1 problem with using NODE_ENV=test. The configurator changes NODE_ENV when we use `--production` flag. And it kills jest

src/bin/commands/setup/__tests__/inquirer.js
thangngoc89: nope. I removed them

thangngoc89:
I gave up on this branch. If you take this over, revert it to `f4b13e4`

MoOx:
I am finishing this.

MoOx:
I don't like having to add a babel env test section in base theme but I will find a way to hide all this config behind a preset before cutting a release.

PR #746, opened 2016-09-14
Changed: better terminal output (for static & server)

### LINKED ISSUE #739: Build doesn't show `phenomic:static` log entries after upgrading to 0.16.0
0.15.0

image

0.16.2

image

MoOx:
Better is to try the dev server yourself. You will enjoy the nice spinner.

Closes #739 

screen shot 2016-09-15 at 00 45 25


docs/package.json
```diff
@@ -16,9 +16,11 @@
     "lint:js": "eslint --ignore-path .gitignore --fix .",
     "lint:css": "stylelint \"src/**/*.css\"",
     "lint": "npm-run-all --parallel lint:*",
+    "rebuild-phenomic": "cd ../ && npm -s run transpile",
```
thangngoc89: Do you think this is the best approach? 

I usually left a terminal tab open which is running this command: 

```
npm run transpile -- --watch
```

Combining this with `npm-run-all` in `start`, we won't have to kill Webpack process to rebuild Phenomic.

src/builder/server.js
```diff
@@ -178,7 +188,7 @@ export default (config) => {
         throw err
       }
       const href = `http://${ devHost }:${ port }${ config.baseUrl.pathname }`
-      log(`Dev server started on ${ href }`)
+      log(`Development listening on ${ href }`)
```
thangngoc89: IMHO, this sounds bad. Maybe `Development server listening on` ...

docs/package.json
MoOx: This was causing me issue when working on cache/dll, maybe I should do that now :)

src/builder/server.js
MoOx: Will fix that.

MoOx:
Should be good. Please review :)

PR #747, opened 2016-09-15
Use stateless component

### LINKED ISSUE #46: Switch to (stateless) functional components
...instead of classes.

thangngoc89:
Fixes #46

### DIFF
--- a/src/components/PageContainer/component.js
+++ b/src/components/PageContainer/component.js
@@ -135,27 +135,26 @@ class PageContainer extends Component<DefaultProps, Props, void> {
       return
     }
 
-    if (this._content) {
-      const layoutDOMElement = findDOMNode(this._content)
-      if (layoutDOMElement) {
-        let bodyContainers =
-          Array.prototype.slice.call(
-            layoutDOMElement.querySelectorAll(".phenomic-BodyContainer")
-          )
-        if (!bodyContainers.length) {
-          bodyContainers = [ layoutDOMElement ]
-        }
-        catchLinks(bodyContainers, (href) => {
-          const pageUrl = href.replace(process.env.PHENOMIC_USER_PATHNAME, "/")
-          if (!find(this.context.collection, pageUrl)) {
-            return false
-          }
-          if (browserHistory) {
-            browserHistory.push(pageUrl)
-          }
-          return true
-        })
+    const layoutDOMElement = findDOMNode(this)
+
+    if (layoutDOMElement) {
+      let bodyContainers =
+        Array.prototype.slice.call(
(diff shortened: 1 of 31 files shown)


src/components/PageContainer/component.js
```diff
@@ -252,7 +252,11 @@ class PageContainer extends Component<DefaultProps, Props, void> {
         return <PageLoading />
       }
       else if (Layout) {
-        return <Layout ref={ this.saveContentRef } { ...page } />
+        return (
+          <div ref={ this.saveContentRef }>
```
MoOx: This is not an option. I recently made a change to remove this div. I know that ref cannot be used on stateless component, but I don't want a div to be added neither.

thangngoc89: I updated this.

thangngoc89:
I updated this PR. Please review

MoOx:
FYI this PR is responsible for breaking hot loading.
The commit before (with webpack@2.1.0-beta.22) works fine, this one does not. I am even having lot's of bug  (docs does not work without changing some files...)
I don't have time to investigate this right now tho, so if you have some times, that would be nice (this regression is blocking me for cutting a release).

MoOx:
Hummm maybe not after all. This is pretty hard to debug, as always...

thangngoc89:
I double checked the diff and I can't find any bugs in this PR. I'll help you debug this once I can get Ubuntu working. Updating Windows 10 broke Ubuntu. F*ck you Microsoft.

MoOx:
It seems that latest react-hot-loader version don't like component using `(props) => (` (returning jsx without `{ return`. By removing that I have an app working correctly. Otherwise lot of issue saying

```
Warning: React.createElement: type should not be null, undefined, boolean, or number. It should be a string (for DOM elements) or a ReactClass (for composite components)
```

Very hard to track. I had to update ALL components :/
Will confirm this and open an issue on RHL repo if relevant

thangngoc89:
Weird. I didnt see any warnings when I sent this PR

MoOx:
I have to admit I have the feeling I have lost 8 hours of my time. Issues seems completely random. That's fucking weird. Will push something later today.

MoOx:
Commit before this PR

```
git checkout 8e0040d8c49ce0c69d8b5883a15102d430f619a4
npm i webpack@2.1.0-beta.22
./node_modules/.bin/babel-node scripts/docs.js
cd docs
npm start
```

Hot loading work.

// this PR

```
git checkout 5935b9b1a8b473c2307ddc8621ded576c9d10ae3 
npm start
```

The app start but as soon as you edit any file an expect hot loading, you got a bunch of errors.

screen shot 2016-09-25 at 21 51 15

PR #749, opened 2016-09-15
Add babel-preset-es2015 to devDependencies

thangngoc89:
No idea how this happened but this break stuff on a clean clone and install


thangngoc89:
I updated the PR. Please review

PR #751, opened 2016-09-16
Next release

MoOx:
I am going to handle some randoms issues before cutting a release, you can follow the progress here :)


src/configurator/validators/offline.js
```diff
@@ -165,9 +165,15 @@ export default (
   }
 
   // Disable offline for development if user defined offline config
-  if (config.dev && config.offlineConfig.appcache) {
+  if (config.dev && config.offline) {
```
thangngoc89: This doesn't disable SW in dev mode if user set `offline = true`

It should be `if(!config.dev && config.offline)`

MoOx:
@thangngoc89 feel free to review

PR #753, opened 2016-09-17
Added: Specify a folder for babel-loader temp dir

thangngoc89:
As we have plan to switch to Jest so I wrote tests that doesn't depend on test runner. They looks cleaner right? 

Also, [the new webpack loader configuration doesn't work](https://gist.github.com/sokra/27b24881210b56bbaff7#loaders-configuration). I tried it in docs and webpack throw errors.  

The gist of edit-webpack-loader-config

Convert all config to webpack 2 `loaders` style. Do the manipulation. Convert it back to webpack 1 `loaders` style if needed.


thangngoc89:
Don't merge this yet. A PR has been sent to babel-loader to add this behaviour. 

https://github.com/babel/babel-loader/pull/301

MoOx:
This had been solved upstream.

PR #757, opened 2016-09-17
Add help for `$ phenomic` and unknown commands

### LINKED ISSUE #513: `$ phenomic` does nothing
We should output something like "use start or build" or something like that.
Or maybe just the help.

thangngoc89:
Handle some small details of CLI: 
- Added usage section
- Added help when typing `$ phenomic` without any command/ options (Closes #513)
- Added help when typing an unknown commands
- Updated yargs to 5.0.0 (Closes #624)


src/configurator/yargs.js
```diff
@@ -6,8 +6,8 @@ import definitions from "./definitions.js"
 
 yargs
   .version(() => version)
+  .usage("Usage: phenomic <command> [options]")
   .help()
-  .showHelpOnFail()
```
MoOx: Is this thing not working?

thangngoc89: It's not working at all since we doesn't do any validation with `yargs.check()`. But this is actually an interesting idea. I could move the validation part to `yarts.check()` and let yargs do the jobs.

thangngoc89:
I'm busy right now. Will do this tonight

MoOx:
Yeah sure, no rush :)

thangngoc89:
I updated the PR.

Some screenshots: 

image

image

src/configurator/yargs.js
```diff
@@ -6,7 +6,8 @@ import definitions from "./definitions.js"
 
 yargs
   .version(() => version)
-  .help()
+  .usage("Usage: phenomic <command> [options]")
+  // .hep()
```
MoOx: oups :)

src/bin/phenomic.js
```diff
@@ -68,4 +69,29 @@ yargs.command(
...
+      colors.bgRed(colors.white("ERROR")) +
+      colors.red(" You must specify command for Phenomic CLI")
+    )
+  }
+
+  // Unknown command
+  const currentCommand = argv._[2]
+  if ([ "start", "build", "setup" ].indexOf(currentCommand) < 0) {
+    throw new Error(
+      colors.bgRed(colors.white("ERROR")) +
```
MoOx: We should probably try to output a more fancy error prefix than "ERROR" :D

src/bin/phenomic.js
```diff
@@ -68,4 +69,29 @@ yargs.command(
...
 )
 
+yargs.check((argv) => {
+  // requires at least 1 argument
+  // show help when type `$ phenomic`
+  // argv._ = [
+  //   nodePath,
+  //   pathToPhenomicScript,
+  //   input argvs
+  // ]
```
MoOx: Can you remove this commented code?

src/configurator/yargs.js
thangngoc89: oops :D

src/bin/phenomic.js
thangngoc89: OK. I added this on purpose to explain about `argv._.length < 3`

src/bin/phenomic.js
thangngoc89: @MoOx what are you suggesting?

thangngoc89:
I'll add some tests with `suppose`

thangngoc89:
Also add `phenomic clean-cache` to delete `./node_modules/.cache/phenomic`. ETA: unknown

MoOx:
Will be handled in #936

PR #758, opened 2016-09-17
Upgrade babel-plugin-flow-react-proptypes to ^0.10.1

thangngoc89:


MoOx:
WTF I changed the base branch and it fucked up everything. Will revert later. Github feature is not working as expected...

thangngoc89:
What fucked up? The docs-update doesn't contains those 2 commits from master, so it's normal. But then you squashed and merged, so this mess. 😂 Github should have warned us

MoOx:
Yeah, I did that too quickly. Mess fixed.

PR #764, opened 2016-09-19
[WIP] Remove restriction on babel-register hook

### LINKED ISSUE #450: Consider using babel-node instead of babel to avoid the *.babel.js need in node context
File consumed via webpack build don't need this, but since we removed babel-plugin-webpack-loaders, I also removed babel-node.
We should try using it back and check perfs. If same result (or almost, since require hook is already on, to handle those *.babel.js via https://github.com/tkellen/node-interpret) we can add it back.

thangngoc89:
Closes #450


MoOx:
By doing that we could remove babel.js extension from docs & boilerplate & tell people to do it right?

thangngoc89:
Yep

montogeek:
Can we merge this?

MoOx:
Tests are red. We should rebase on master, then fix the build :)
I guess you can continue iterating on this branch since you have the right to!

Issue #621, opened 2016-08-16
Generating a sitemap


DavidWells:
How would you suggest handling XML sitemap generation?

There seem to be a lot of NPM packages for this https://www.npmjs.com/search?q=sitemap+generator 

Does anyone have a recommended way of doing this?

thangngoc89:
You have all the information you need in collection to create a sitemap. Just need to print it out properly. Or you can just use the RSS feed generating feature. With Google Webmaster tool, you can submit a RSS feed for sitemap.

MoOx:
I guess we can add this like for the rss feed, and later extract this as a package when handling #598.
This would have to be added here https://github.com/MoOx/phenomic/blob/2b44e19bfa095d2a4bb512c122dbb24380a83ef1/src/phenomic-loader/index.js#L86

MoOx:
@thangngoc89 having informations is one thing, being able to do it without pain is another, we need to provide options to make this super easy. It's very a basic requirement.

thangngoc89:
I suggest extracting the rss and sitemap logic from the loader and put it into the plugin.

MoOx:
baby steps ;) For now it's easy to add this here.

theduke:
So, how do I go about adding a sitemap?

The problem I see is:

```javascript
<Route component={ LayoutContainer }>
    <Route path="/sitemap.xml" component={SiteMap} />
    <Route path="*" component={ PageContainer } />
</Route>
```

So how do I define an additional "/sitemap.xml" route that does not use LayoutContainer?

MoOx:
In theory, this should work (for any html url), but the problem is that pages generated are wrapped into an Html component.
My suggestion is to use something else and not rely on phenomic (eg: https://www.npmjs.com/package/simple-sitemap)

theduke:
But I need the sitemap to be available at /sitemap.xml, not just submit it to google.

I guess I could write a script to generate it and copy it to the build folder on deploy though.

MoOx:
The package I mentionned allow to write to disk, so yeah you should call it after the build (eg: npm "postbuild" hook) and this should do the trick.

thangngoc89:
This should be easy now, but still require some works.

You can take a look at [Loader Feed Webpack Plugin](https://github.com/MoOx/phenomic/blob/master/src/loader-feed-webpack-plugin/index.js) and make a similiar Plugin (Loader Sitemap Webpack Plugin)

[...]

Issue #142, opened 2016-02-14
Add Search capabilities


MoOx:
Should be very easy since we have collections to search in all meta (title, tags etc) - but not in body of the posts.
Maybe agolia can help for the body (ref https://github.com/putaindecode/putaindecode.io/issues/535)

thangngoc89:
This is interesting. We can make a different md-collection-loader for generating data and feed it with algolia

MoOx:
Yeah an option like for rss feed.

thangngoc89:
How about a different loader or maybe a new package (like a plugin) ?

thangngoc89:
I think this should not be in the core. This belongs to a plugin. Already done this. See https://github.com/thangngoc89/blog/blob/aa40e1b710c06112006ac175972338467adb56b3/web_modules/utils/statinamic-agolia-plugin.js

Generate feed should be done this way too. No need for debounce trick (faster markdown build)

thangngoc89:
We can borrow idea from https://blog.algolia.com/instant-search-blog-documentation-jekyll-plugin/

MoOx:
Also interesting http://lunrjs.com/

thangngoc89:
This is what it takes to add agolia to a Statinamic site

https://github.com/thangngoc89/dnh-blog/commit/21b07d842a3fd40a5ee362483ba1906cfcd09b55

Demo: http://blog.daynhauhoc.com/

MoOx:
ETA?

thangngoc89:
Lunr.js or algolia ?  Or a simple component that search collection ?

MoOx:
We might offer the choice during setup and/or have the simplest solution (without third-party) and offer a doc to add another solution.

thangngoc89:
I'm on it. 

# TODO:

- [ ] Build index
- [ ] Reduce api requests to Algolia on each build by hashing the page content and only send changed page.

# Resources: 

- https://blog.algolia.com/instant-search-blog-documentation-jekyll-plugin/
- https://blog.algolia.com/how-to-build-a-helpful-search-for-technical-documentation-the-laravel-example/

# Example record 

```
{
    "h1": "Validation", 
    "h2": "Working With Error Messages", 
    "h3": "Custom Error Messages", 
    "link": "validation#custom-error-messages", 
[...]

PR #768, opened 2016-09-20
[WIP] Add search function with Algolia

thangngoc89:
This PR is far from complete, but I want to get early feedback.
- [x] Build index
- [ ] Publish index to Algolia
- [ ] Lazy update
- [ ] Front-end implementation

@MoOx @DavidWells 

If you clone this branch and build docs, you'll get a `search-index.json`. We can upload it to algolia to create an index. I dumped the content of [`search-index.json` here](https://gist.github.com/thangngoc89/727dc50ede6cbc8c7f7381ea7705132b)

Online demo: https://www.algolia.com/realtime-search-demo/phenomic-docs-search

### DIFF
--- a/src/_utils/search/create-index/webpack-plugin.js
+++ b/src/_utils/search/create-index/webpack-plugin.js
@@ -3,7 +3,7 @@ import { RawSource } from "webpack-sources"
 import PhenomicLoaderWebpackPlugin from "../../../loader/plugin.js"
 import createIndex from "./index"
 
-function CreateSearchIndexWebpackPlugin(options: Object) {
+function CreateSearchIndexWebpackPlugin(options?: Object) {
   this.options = options
 }
 
--- a/src/_utils/search/create-index/webpack-plugin.js
+++ b/src/_utils/search/create-index/webpack-plugin.js
@@ -1,0 +1,42 @@
+// @flow
+import { RawSource } from "webpack-sources"
+import PhenomicLoaderWebpackPlugin from "../../../loader/plugin.js"
+import createIndex from "./index"
+
+function CreateSearchIndexWebpackPlugin(options: Object) {
+  this.options = options
+}
+
+CreateSearchIndexWebpackPlugin.prototype.apply = function(compiler) {
+  compiler.plugin("compilation", (compilation/* , params */) => {
+    compilation.plugin("additional-assets", (callback) => {
+      if (!PhenomicLoaderWebpackPlugin.collection) {
+        throw new Error(
+          "Missing Phenomic collection in webpack compilation object. " +
+          "This probably means you are playing with fire."
+        )
+      }
+      const collection = PhenomicLoaderWebpackPlugin.collection
+      const searchIndex = []
+
+      collection.forEach((item: PhenomicCollectionItem) => {
+        const hierarchy = createIndex(item.body)
+        hierarchy.forEach((p, index) => {
+          searchIndex.push({
+            objectId: item.__url + "#" + index,
+            title: item.head.title || item.head.metaTitle,
+            __url: item.__url,
+            __dataUrl: item.__dataUrl,
+            ...p,
+          })
+        })
+      })
+      compilation.assets["search-index.json"] = new RawSource(
+        JSON.stringify(searchIndex, null, 2)
+      )
+      callback()
+    })
+  })
+}
+
+module.exports = CreateSearchIndexWebpackPlugin
--- a/docs/webpack.config.babel.js
+++ b/docs/webpack.config.babel.js
@@ -87,7 +87,7 @@ export const makeConfig = (config = {}) => {
     },
 
     postcss: () => [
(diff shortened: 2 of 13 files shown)


src/_utils/search/create-index/webpack-plugin.js
```diff
@@ -0,0 +1,42 @@
...
+CreateSearchIndexWebpackPlugin.prototype.apply = function(compiler) {
+  compiler.plugin("compilation", (compilation/* , params */) => {
+    compilation.plugin("additional-assets", (callback) => {
+      if (!PhenomicLoaderWebpackPlugin.collection) {
+        throw new Error(
+          "Missing Phenomic collection in webpack compilation object. " +
+          "This probably means you are playing with fire."
+        )
+      }
+      const collection = PhenomicLoaderWebpackPlugin.collection
```
DavidWells: Woah cool. Is this how you can access the entire collection via webpack?

MoOx: Yeah (currently), I hope to offer a better API soon. That's what @thangngoc89 pointed out here https://github.com/MoOx/phenomic/issues/621#issuecomment-247192443

rayrutjes:
Looks good @thangngoc89 
Then it will mainly come down to the index configuration as well.

Also, when it comes to splitting up content based on the DOM structure, there are two strategies.

**Minimum amount of records**

Will avoid having a record with an H1 alone in it in favor of records going always as deep as possible in the hierarchy

**Verbose**

Will create a record for each level in the DOM structure


The second option allows you to display result suggestions for a same page to let users access a specific anchor of that page. Will result in a lot more records though.

The choosing between option 1 and 2 comes down to the type of site. For a documentation you probably want users to be able to jump right into a section of a page. For a blog or a corporate website you probably just want people to be heading to the top of the page.

Hope that helps, I'd love to find some time giving your solution a try, looks gorgeous!

vvo:
> Then it will mainly come down to the index configuration as well.

This is doable all via the API, using https://github.com/algolia/algoliasearch-client-js

thangngoc89:
@vvo thank you. That's what I'm looking for. 

@rayrutjes so you're saying that if I don't need to **let user users access a specific anchor of the page** . I can just use 1 record for each post? What about the 10KB per records?

rayrutjes:
>  so you're saying that if I don't need to let user users access a specific anchor of the page . I can just use 1 record for each post?

Not exactly because if you choose to index 1 record per post you will reach the 10kb limit. Actually we even recommend you stay under 3kb ;)

What I'm saying is that in documentation websites, it makes sense to have a record pointing toward an H1 anchor of a page, so having a record containing only the H1 makes sense there.

In non documentation website, you don't need those records because you are only interested in global relevancy of the result. So having the H1 part of another record containing the children is OK.

When splitting posts/pages into multiple records, you can use the DISTINCT feature of Algolia to only get back the most relevant record for a given post/page.

thangngoc89:
@rayrutjes thank you. I got it. I think I'll go with docsearch-configs `https://github.com/algolia/docsearch-configs`. It's pretty flexible and user can choose which selector they want to index. I'm learning some python to see how [docsearch-crawler](https://github.com/algolia/docsearch-scraper) works

MoOx:
@thangngoc89 I guess docsearch make sense for #769, maybe not for all websites, right?

Thanks @rayrutjes @vvo for your inputs, it's really nice from you :)

vvo:
> I guess docsearch make sense for #769, maybe not for all websites, right?

Yes it makes sense only for Phenomic website docs, we will provide a free crawling and indexing documentation infrastructure like we do for react/react-native..

At least not the docsearch "SAAS", where Algolia company configures, host for free your index. We do provide free DocSearch installs even for websites where the free plan would not fit. If the website is compatible with what we call community websites, websites that are aimed at developer communities.

But if $BIGCORP uses Phenomic then we won't provide a free DocSearch install for instance.

Still, you could take the DocSearch crawler and index format to achieve what we do yes. That may be a bit overengineered for a JavaScript project. Maybe also you do not want to maintain some python programs.

So a simple approach could be to just do a small index like you did @thangngoc89 and use react-instantsearch (in alpha, @thangngoc89 has it), to provide a simple set of components inside Phenomic.

thangngoc89:
I'll keep the surface of the PR small, I won't change the index builder algorithm in this PR. I'll focus on completing all of those tasks above. We can add more complex algorithm and options to the index builder later.

Also, it's a good time for a monorepo because I see other Node.js SSGs can use what I have written here to implement their index builder for Algolia. 

Furthermore, Phenomic will has 2 more commands: 

```
# set index config to Phenomic's recipe
phenomic search algolia:config --masterKey (or whatever)

# publish index to Algolia 
phenomic search algolia:publish (lazy update option will come later) 
```

MoOx:
Closing as this PR have to be completely rewrote (new engine, API etc). #142 is still open.

PR #769, opened 2016-09-20
Docs: Add Algolia DocSearch

thangngoc89:
I registered Phenomic to [Algolia DocSearch](https://community.algolia.com/docsearch/), after a few emails with the team, a search engine for us is ready.

I implemented the logic, you can see it in action here: https://phenomic-docsearch.surge.sh

The styling needs a little love and I'm bad at it. Could you do the styling ?


DavidWells:
@thangngoc89 do you prefer this drop in algolia setup vs the setup you rolled here: https://github.com/daynhauhoc/blog/tree/master/data ?

thangngoc89:
@DavidWells if my site can be submited to DocSearch, I prefer it. Because it's not just about writing the code, you have to configure the index, keep an eye on the API usage, maintenance,... WIth DocSearch, all of this is took care by Algolia team.

If I need to roll my own, I'm working on integrating it https://github.com/MoOx/phenomic/pull/768 .

MoOx:
The idea of #142 is to provide a way for people to easily use Algolia and not just docsearch, so we should have that in mind when handling this issue. I will review this later as I have a fucking lot to cleanup for webpack2 beta3 mess :)

thangngoc89:
@MoOx yes. Please refer #768 for #142 . This is for us (and others OSS projects like cssnano or serverless, stylelint, postcss). cc @ben-eb @-jeddy3 @-marcustisater

MoOx:
Thanks for the explanation. Will take a look to this one right after the release then.

PR #776, opened 2016-09-22
Re: 766 Fix Flow 0.32.0 annotations

montogeek:
Implements #716

### DIFF
--- a/src/builder/webpack/config.common.js
+++ b/src/builder/webpack/config.common.js
@@ -18,10 +18,12 @@ export default (config: PhenomicConfig): WebpackConfig => {
   const cacheDir = getCacheDir(config)
   const { webpackConfig = {} } = config
 
-  const defaultConfig = {
+  return {
     ...webpackConfig,
+    ...config.cache ? hardSourceRecordsPath(cacheDir) : {},
     plugins: [
       ...webpackConfig.plugins,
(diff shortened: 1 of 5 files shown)


src/builder/webpack/config.common.js
```diff
@@ -41,4 +38,17 @@ export default (config: PhenomicConfig): WebpackConfig => {
...
+      ...defaultConfig,
+      ...hardSourceRecordsPath(cacheDir),
+      plugins: [
+        ...defaultConfig.plugins,
+        hardSourcePlugin(cacheDir, config),
+      ],
+    }
+  }
+
+  return defaultConfig
```
MoOx: Can you avoid this big block and instead do something like `...config.cache ? hardSourceRecordsPath(cacheDir) : {},`?

montogeek: Yes, we can!

MoOx:
You work is good @montogeek. It's ANOTHER breaking change in webpack 2 latest beta.

Thanks for the effort!

montogeek:
OMG! Thanks! :D

MoOx:
Actually I was not careful enough, this bug was introduced by this PR :) Anyway we detected an improved error handling.

montogeek:
What bug?

PR #782, opened 2016-09-25
Remove markdown images from appearing in description text

DavidWells:
Markdown images and italics and stuff are showing up as plain text in the collection description.

image

image

These regex added removes markdown images from text before it's processed

### DIFF
--- a/src/loader-plugin-markdown-init-head.description-property-from-content/index.js
+++ b/src/loader-plugin-markdown-init-head.description-property-from-content/index.js
@@ -11,7 +11,9 @@ const defaultOpts = {
 
 function description(text, opts = {}) {
   opts = { ...defaultOpts, ...opts }
-
+  // remove image tags from markdown
+  text = text.replace(/\!\[.*?\)/g, '')
+  
   if (opts.pruneLength === 0) {
     console.warn(
       "You defined 'description.pruneLength' of phenomic loader " +
@@ -21,7 +23,7 @@ function description(text, opts = {}) {
 
     opts.pruneLength = defaultOpts.pruneLength
   }
-
+  
   return prune(
     remark()
       .use(stripMd)


DavidWells:
image

DavidWells:
Here is an example of italics sneaking through:

image

thangngoc89:
Printing image's description is a feature of strip markdown https://github.com/wooorm/strip-markdown/blob/77cc294226fe09ada6cee761a01b5d6839ba54a5/test.js#L104 

The italic text looks like a bug. But I prefer sending a PR/open an issue to [strip-markdown](https://github.com/wooorm/strip-markdown)

DavidWells:
@thangngoc89 just double checked. The italics was actually my fault =P

But the image thing is something I think we should remove. The left-over image text makes the snippet read incorrectly

DavidWells:
@moox @thangngoc89 does this need to be done upstream at the strip-markdown dependancy? 

I can't really think of a reason why someone would want "image description text" mixed in with actual text.

DavidWells:
You see?

image

thangngoc89:
@DavidWells well, create a loader plugin that run the regex before description plugin. You can totally do this.

MoOx:
Will be in core. But we should add a test for this.

DavidWells:
Now that I look at the tests. I'm wondering if "plain-texting" the description is ideal.

https://github.com/DavidWells/phenomic/blob/master/src/loader-plugin-markdown-init-head.description-property-from-content/__tests__/index.js#L50

```
# Why I love apples 

Oranges are great but apples are better... This is my paragraph that says awesome stuff. 
```

Turns into: "Why I love apples Oranges are great but apples are better... This is my paragraph that says awesome stuff." (bad grammar, improper english)

Aside from this example blog post being ridiculous... Do you think we should perhaps exclude headings/lists/tables etc from automatic descriptions? 

Proposal:

- strip some default tags (headings/lists/tables) from automated description but add option for people to include them?

DavidWells:
Then again. This would be solved if people do their job of adding the 'description' frontmatter =P

I'm conflicted

MoOx:
Automatic description is a nice feature imo. At lease for meta tags.
Stripping eveything that is not a paragraph make sense to me.
If people need a custom behavior -> custom plugins.

thangngoc89:
If you want to remove heading, lists, tables, go with remark plugin, add a new remark plugin that strip all of those nodes before strip markdown. 

@MoOx I see that we are parsing markdown twice with remark. Is it a good idea to merge generate body and description into 1?

wooorm:
Does strip-markdown need to change for this, and if so, how?

wooorm:
I haven’t touched that for a while. I should really clean that codebase.

MoOx:
Maybe this should be an option in strip-markdown indeed! Something like an "agressive" flag?

wooorm:
##### Images

Why is there image alt text if you don’t want people seeing/hearing/reading it?

##### Headings

The fact that

```md
# Why I love apples 

Oranges are great but apples are better... This is my paragraph that says awesome stuff. 
```

Turns into...

```txt
Why I love apples  Oranges are great but apples are better... This is my paragraph that says awesome stuff. 
```

Is not done by strip-markdown, I’m guessing it’s handled in phenomic somewhere?

thangngoc89:
@wooorm we rely on strip markdown for generating description from main content. See https://github.com/MoOx/phenomic/blob/2f46be51f1fdab5bc2d07323ce427faf1398cb46/src/loader-plugin-markdown-init-head.description-property-from-content/index.js#L26-L32 

We don't have any custom behaviours here.

wooorm:
[This behaviour](https://github.com/MoOx/phenomic/blob/2f46be51f1fdab5bc2d07323ce427faf1398cb46/src/loader-plugin-markdown-init-head.description-property-from-content/index.js#L30-L31) is what I’m talking about, that’s done here.

MoOx:
Indeed, the lines replacing \n are a custom behavior weird enough to make the output weird.

thangngoc89:
But since the description is intended to be used in meta tag, having multiple lines doesn't make any sense.

MoOx:
Yeah, but this is causing issue in this case. Pretty tricky right?
Not sure what is the best to do. Probably stripping some elements like we discussed which will make sense in some context but not when inlined.

thangngoc89:
This is a matter of opinion and use cases. You may need both in the same project.

MoOx:
I don't see how someone will need an inlined text like this: "Title alt_image Hi there this is a new post..."

thangngoc89:
> I don't see how someone will need an inlined text like this: "Title alt_image Hi there this is a new post..."

On that I agree

MoOx:
Up here. Should we remove everything that is not just normal text for the desc?

MoOx:
I am closing this PR as for now
- transform markdown plugin (in master) does not create a description by default as many people where not happy with it (and even confused)
- the current PR is doing a change in a file that does not exist.

We will revisit this later if we reimplement this, but I guess for now this should be handled in user land.

PR #785, opened 2016-09-26
0.17.0

MoOx:


thangngoc89:
Finally, it arrived. You may need to clean appveyor cache

MoOx:
Nope there is a real bug but no big deal. Writing release notes with some bench right now.

PR #789, opened 2016-09-28
The windows documentation should use backslashes for path

maxdow:
Windows still uses backslashes for path


MoOx:
Oops :) Thank you!

PR #794, opened 2016-09-29
Fixed: assets=fasle breaks offlinePlugin

thangngoc89:
Assets option can be `false` which breaks offlinePlugin.

### DIFF
--- a/src/configurator/validators/assets.js
+++ b/src/configurator/validators/assets.js
@@ -15,7 +15,9 @@ export default function(
   if (config.assets) {
 
     // normalize simple string options
-    if (typeof config.assets === "function") {
+    if (
+      [ "boolean", "string", "object" ].indexOf(typeof config.assets) < 0
+    ) {
       errors.push(
         "You provided an function for 'assets' option." +
         "This option accept a boolean value, a string, or an object."
@@ -28,6 +30,7 @@ export default function(
           typeof config.assets.route !== "string"
         )
     ) {
+      const configAssets: PhenomicAssetsConfig = config.assets
       errors.push(
         "You provided an object for 'assets' option." +
         "You need to provide 2 keys: " +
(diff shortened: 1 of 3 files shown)


thangngoc89:
@MoOx could you review this? And maybe cut a release

MoOx:
Will do later today, I have a huge fix ready for offline. Stay tuned.

MoOx:
It's weird that flow didn't find this issue earlier. Or maybe our config is not correctly formed

MoOx:
The proper fix should start by doing something like we did for offline /offlineConfig

https://github.com/MoOx/phenomic/blob/36b03832dde5cf3a11d1b3480f15935172697060/flow/interfaces/phenomic.js#L20

We should validate "assets", then create and assetsConfig.

Or, alternatively, we could just assume during validation it's correctly formated OR set to false, so we could adjust this

```js
export type PhenomicAssetsConfig = {
  route: string,
  path: string,
}
```

to

```js
export type PhenomicAssetsConfig = boolean | {
  route: string,
  path: string,
}
```

thangngoc89:
PR updated. Please review

src/configurator/validators/assets.js
```diff
@@ -15,7 +15,10 @@ export default function(
   if (config.assets) {
 
     // normalize simple string options
-    if (typeof config.assets === "function") {
+    if (
+      typeof config.assets === "function" ||
+      typeof config.assets === "number"
```
MoOx: maybe more: != object && != boolean?

thangngoc89: yeah. that's better. Thanks

src/configurator/validators/assets.js
```diff
@@ -55,10 +59,11 @@ export default function(
         }
       }
 
+      const configAssets: PhenomicAssetsConfig = config.assets
```
MoOx: maybe we can avoid this repeatition?

thangngoc89: We can't assign `configAssets` at the beginning of that function.Because we adjust `configAssets` above this line. Flow doesn't understand the situation here.

src/configurator/validators/assets.js
thangngoc89: I adjusted this.

src/configurator/validators/assets.js
```diff
@@ -15,7 +15,9 @@ export default function(
   if (config.assets) {
 
     // normalize simple string options
-    if (typeof config.assets === "function") {
+    if (
+      [ "boolean", "string", "object" ].indexOf(typeof config.assets) > -1
```
MoOx: <3

PR #797, opened 2016-09-30
Fix commented out example for global CSS to be valid JS syntax

ernieturner:
Noticed that when trying to use this commented out code sample that it wasn't fully valid JS. I also noticed that it's using Webpack2 syntax, so if you'd like I can add that note in there too (since it seemed that most Webpack2 specific stuff was commented that way).


MoOx:
Thanks!

MoOx:
I should add a note indeed :)

PR #798, opened 2016-09-30
keep URL params

DavidWells:
Before: http://localhost:3333/partners?jhjajhdsa=yo#lol automatically removes query + hash -> http://localhost:3333/partners/

After: URL is fixed (trailing slash added) and query string + hash left in place http://localhost:3333/partners?jhjajhdsa=yo#lol

This makes it possible for links with hashes to retain hashes.

It also makes it possible for us to use query params in our site. EG utm tracking for marketing


MoOx:
Looks good. Thanks a lot!

PR #804, opened 2016-10-03
Create www.agrista.com

drewes:
Would like to submit our new website to the showcase.

Thanks for the framework. It also was a great framework for understanding React.


MoOx:
Nice website! Thanks :)

PR #807, opened 2016-10-05
updated eslint to version 3.7.1

comoss:
I updated the package.json to reference the latest version of eslint, let me know if there are anymore references that should be updated.


MoOx:
Hi @comoss. Thanks a lot for your contribution :)
I guess it's also time to update eslint-* packages to latest version. Like eslint-config-i-am-meticulous & friends. This may involve adding eslint-plugin-import :)

MoOx:
And now it's failing from everywhere because of import new rules :D
Maybe a simple addition (until we adjust the entire codebase) will be to add "eslintConfig" to disable some import rules that are causing issues.

comoss:
that sounds good, however I am a little confused as to how to implement that. I see that there already is a eslintConfig section in the package.json. Should I be disabling the rules there? and which rules should be disabled?

MoOx:
yeah you should add a "rules" section under the eslintConfig section.
See here to get the full list of rules to disable https://travis-ci.org/MoOx/phenomic/jobs/165138879#L321

Just disable all rules for now, I might handle that later :)

Again thanks for your work.

MoOx:
You didn't take the easiest issue :D
You can probably disable more rules for now until it passes. Maybe you can run eslint locally so you can do this in one shot & not waiting for the CI to pass or fail?

comoss:
Yeah I can try that, sorry!!

MoOx:
No problem. I am just saying that so you can save time :)

MoOx:
You will also need to add this in the package.json template, right here https://github.com/MoOx/phenomic/blob/0155cb103f0decfa13a2deb5fc3f04cbfeafb196/src/bin/commands/setup/template.js#L75

comoss:
@MoOx do you know what could be causing the appveyor to fail? I tried looking at the log, but I didn't get much out of it.

MoOx:
Don't worry, it's not your fault. It was already like that. Nice work!
I will merge in another branch since I want to change some rules and things, but you already did a great work :)

Issue #818, opened 2016-10-11
Support Yarn


MoOx:
https://github.com/yarnpkg/yarn

Yarn is a faster alternative to the npm client. 
- We should be able to detect when yarn is installed for faster testing (since we do some clean install)
- We should provide command for yarn users (I guess)

dflynn15:
✋ 🙇

MoOx:
This might be super interesting https://github.com/javascriptair/site/blob/master/.travis.yml (we can thanks @kentcdodds)

DavidWells:
Good overview of yarn for folks interested https://laracasts.com/lessons/meet-yarn-your-new-package-manager

daaain:
FYI the current setup process right now fails when using Yarn, see #869. There's a Dockerfile template in the issue for testing if you want.

PR #823, opened 2016-10-12
Adds yarn and documentation

dflynn15:
- Adds `yarn` lock file
- Adds `yarn` to documentation
- In reference to #818 

Let me know if I missed anything!!
💻 🐈 💨 ✨


MoOx:
@dflynn15 Thanks a lot for you contribution. I really appreciate.
One thing I would like to do is not to drop npm support (yet?) So I think we should have doc for both.
Also I notice that you replaced all occurence of npm, but not all command are supported:
- npm scripts should still be npm script (eg: npm start, npm run etc)
- npm install -> yarn
- npm install X -> yarn add

One thing we should do is: use yarn for travis CI (should boost test) but keep npm on appveyor (not sure if we can do this easily). Maybe for a start add a flag to "npm test" to be able to use yarn install or npm.

That's more work I know... :D

dflynn15:
Yep! I took the liberal approach and figured it would be easier to reign it back in after review 👍  I'll go back and make those changes!

dflynn15:
@MoOx I'm still working through the Travis build (hence build fails) but I wanted to let you get 👀  on the documentation while I worked on it 👍

MoOx:
Some places still replace npm by yarn when they should not :)

MoOx:
@dflynn15 need some help?

dflynn15:
@MoOx I think I finished up the docs. I am still trying to debug the Travis CI failures though 😭

MoOx:
You will also need to adjust this script https://github.com/MoOx/phenomic/blob/master/scripts/utils/test-folder.js (and maybe we should detect that yarn exist before using yarn here, so the full tests can work with npm or yarn).

MoOx:
@dflynn15 looks better.
I see that travis is reporting a warning pointing to this https://docs.travis-ci.com/user/languages/javascript-with-nodejs#Node.js-v4-(or-io.js-v3)-compiler-requirements

Can you try adding this things? I looked in yarn issues and didn't find anything mentionning the error we are facing. Will try to take a look later if I got some times.

Thanks for your effort!

MoOx:
Trying to make this work on a yarn branch using this PR. Thanks for the work @dflynn15!

dflynn15:
Sorry I couldn't get it finished!! Glad you were able to fix the scripts in a way to get it working 👍

MoOx:
Don't be sorry, you helped a lot!

PR #829, opened 2016-10-19
Handle .markdown files

cedricdelpoux:
When I tried to use `SiteLeaf` I faced to a problem. `SiteLeaf` generates `.markdown` files instead of `.md` files.

So This PR fix that


MoOx:
Awesome! Thank you. Will cut a release this evening.

PR #831, opened 2016-10-19
Update getting started docs

indiesquidge:
### Changes Include:

- removing whitespace from EOL
- fixing URL paths
- updating sentence structure
- including note about the collections

Happy to talk about the changes I have made if they do not make sense!

Thanks for building an awesome platform :)

### DIFF
--- a/docs/content/docs/getting-started.md
+++ b/docs/content/docs/getting-started.md
@@ -24,14 +24,14 @@ Then add into it some content:
 title: My super new page
 ---
 
-My super **content** !
+My super **content**!
 ```
 
 This file is composed of two sections:
 
 ### The front-matter
 
-The first part (between the 3 dashes ``---``) contains page metadata.  
+The first part (between the 3 dashes ``---``) contains page metadata.
 You can store anything you want.
 It's by default a
 [YAML section](https://en.wikipedia.org/wiki/YAML#Sample_document),
@@ -40,7 +40,7 @@ but you can use
 
 ### The body
 
-The body is the part below the 3 dashes ``---``.  
+The body is the part below the 3 dashes ``---``.
 By default [Markdown](https://en.wikipedia.org/wiki/Markdown#Example) is supported but
 [you can easily add whatever format you want](../usage/plugins/).
 
@@ -73,14 +73,14 @@ No problem, you can still add a ``route`` field in the front-matter like this:
 ```md
 ---
 title: My super new page
-route: my-super-url.html
+route: my-super-url
 ---
 
-My super **content** !
+My super **content**!
 ```
 
 Then your content will be accessible at
-[http://localhost:3333/my-super-page.html](http://localhost:3333/my-super-page.html)
+[http://localhost:3333/my-super-url/](http://localhost:3333/my-super-url/)
 
 ---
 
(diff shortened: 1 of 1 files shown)


docs/content/docs/getting-started.md
```diff
@@ -73,14 +73,14 @@ No problem, you can still add a ``route`` field in the front-matter like this:
 ```md
 ---
 title: My super new page
-route: my-super-url.html
+route: my-super-url
```
indiesquidge: Looks as though putting an `.html` on the route breaks if that is not the filetype, whereas without the extension it works for both HTML and Markdown pages.

docs/content/docs/getting-started.md
```diff
@@ -140,16 +140,16 @@ you want!
 We will need to create a new component and references it in some places.
 
 First you need to tell Phenomic that you added a layout.
-So you need to register your layout in ``src/app/routes.js`` in the ``PhenomicPageContainer`` usage.
```
indiesquidge: Doesn't seem there is an `app` directory anymore.

docs/content/docs/getting-started.md
```diff
@@ -197,25 +197,35 @@ layout: MySuperLayout
...
 
 Changing the ``PagePreview`` component will change how the "latest posts" are rendered on your homepage.
 
-```js
+---
+
+⚠️ The `Homepage` layout filters the latest posts by `Post` layout. If you don't
+see your new page in this list, either change your new page to have the `Post`
+layout, or edit the homepage collections filter to include your custom layout
+by checking out the [How to use Collections](https://phenomic.io/docs/usage/collections/) page.
```
indiesquidge: Maybe this is too verbose, but I thought it was hard to follow the example new page we were setting up after we had created a custom layout. I didn't understand why my new page wasn't showing up in the list of posts on the homepage until I saw the collections filter.

MoOx:
LGTM, thank you. I will revert whitespace for EOL as 2 spaces at the end of a line is to generate a new line (via `<br>`).

docs/content/docs/getting-started.md
MoOx: 👍

docs/content/docs/getting-started.md
indiesquidge: @MoOx, apologies for not being clear on this change.

At this point in the tutorial we are working with a file under `content/my-super-page.md`. The contents of that file are, at this point,
```md
---
title: My super new page
route: my-super-url.html
---

My super **content**!
```

Adding the custom route property with the `.html` extension in the front-matter is what is troubling. We are working within a markdown file, and I thought only markdown was supported by default with Phenomic.

With the `.html` extension, I am never able to get the page to load; I only ever see the 404 page. If I remove the extension, the content is available (at `http://localhost:3333/my-super-url/`); if I change the extension to be `.md`, the content is also available (at the same URL). Even changing the filetype to actually be an HTML file does not appear to work. Perhaps I am just not understanding what route the page should now be on?

Either way, it still feels jarring to me introducing the `.html` extension here, since we are never actually using or editing any HTML files in this tutorial.

I would love to hear your thoughts on this though. Again, I could just be overlooking or misunderstanding something about this functionality.

Thanks for taking the time to read this :)

indiesquidge: Ah, yes I see now, I was on version `0.17.6`. All good then! :)

Issue #424, opened 2016-04-27
Online demo boilerplate


MoOx:
It would be nice to have the default boilerplate online.

thangngoc89:
fyi: https://phenomic.surge.sh

I think it's a good idea to use another hosting service to deploy boilerplate.

MoOx:
We can just put it under /demo/
What's wrong with that?

thangngoc89:
Nothing. But it make build step more complicated. Maybe a subdomain ?

MoOx:
No, it won't be complicated at all, don't worry ;)

MoOx:
demo.phenomic.io or phenomic.io/demo?
Wait stop. What about this:

themes.phenomic.io/base or phenomic.io/themes/base?

I think phenomic.io/themes make more sense as we might offers an official "catalog" one day or another?

thangngoc89:
See http://themes.gohugo.io/ if you need inspirations. I think we should go with Hugo's way

DavidWells:
`phenomic.io/themes/base` looks prettier to me but might be harder to maintain.

The subdomain approach might be easier in the long run.

You could have all the themes live on the subdomain and still have a phenomic.io/themes/ page that acts like http://demo.mythemeshop.com/s/?theme=myPortfolio (Not endorsing the design here). It's just a simple iframe that loads in a different URL when you choose a different theme.

MoOx:
We now have the base theme here https://phenomic.io/themes/base/demo/
The way it's deployed is probably temporary until we make a gallery or something like that.

PR #836, opened 2016-10-22
[WIP] update PageContainer

DavidWells:
Code is super dirty but the proof of concept is there!

http://www.screencast.com/t/Ym1KVwFP

Demo here: http://holy-shit-thats-fast.surge.sh/

Going to [/blog](http://holy-shit-thats-fast.surge.sh/blog) and clicking into the different posts will show you the loading states demo.

Before: https://serverless.com/blog/
After: http://holy-shit-thats-fast.surge.sh/blog

### DIFF
--- a/src/components/PageContainer/component.js
+++ b/src/components/PageContainer/component.js
@@ -82,15 +82,7 @@ function getLayout(
   layout: string, props: Props
 ): ReactClass<any> | void {
   if (props.layouts && props.layouts[layout]) {
-    console.log(`props.layouts ${layout}`, props.layouts)
-    const checkForStatic = props.layouts[layout].loadingState
-    if(checkForStatic) {
-      console.log(`layout ${layout} has loading state`)
-    }
-    return {
-      hasStaticLoadingStateSet: props.layouts[layout].loadingState,
-      layout: props.layouts[layout]
-    }
+    return props.layouts[layout]
   }
 }
 
@@ -209,21 +201,16 @@ class PageContainer extends Component<DefaultProps, Props, void> {
     const { props } = this
 
     const pageUrl = splatToUrl(props.params.splat)
-    console.log('pageUrl', pageUrl)
+    // page url from redux store
     const page = props.pages[pageUrl]
-    let actualPage
-    let actualPageType
+    let pageType = (page) ? page.type : null
+    // SSR window check
     if (typeof window !== 'undefined') {
       // use window collection instead of page props
-      actualPage = window.__COLLECTION__.filter(function(pg, i) {
-        return pageUrl === pg.__url
+      const pageFromCollection = window.__COLLECTION__.filter(function(pageData, i) {
+        return pageUrl === pageData.__url
       })
-      actualPageType = actualPage[0].layout
-      console.log('actualPage', actualPage)
-      console.log('actualPageType', actualPageType)
-      console.log('pagepagepagepage', page)
-    } else {
-      actualPageType = page.type
+      pageType = pageFromCollection[0].layout
     }
     if (!page) {
       if (isDevelopmentClient()) {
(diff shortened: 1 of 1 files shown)


MoOx:
Nice concept. I think we can make the code even more simpler. Will have a look as soon as I can.

DavidWells:
Yeah not set on the API or my implementation. The code is messy as is. Just wanted to test it out.

I think keeping the default `<PageLoading>` in place will be nice people and if you want to override the default you can set a static property `static [yournamehere]Loading = true` to do so. 

This lets me not worry about this for every route and allows me to optimize where I want.

MoOx:
I think we can come up with a mix of both solution ;)

DavidWells:
Cleaned up the code and added logging statements.

This should be ready to go, the only thing we need to decide on is the name of the static property to check for. It is currently `loadingState`

```js
export default class MyPage extends Component {
  static loadingState = true // if present this will load instead of generic loading component
  static propTypes = {
    phenomicLoading: PropTypes.bool
  }
  render () {
    const text = (this.props.isLoading) ? 'Loading' : 'Actual Content'
    return (
      <div>{text}</div>
    )
  }
}
```

And also what prop to set to signify loading. `phenomicLoading` or `isLoading`. Do we want it phenomic specific to avoid clashes? or can we make it simply `isLoading`?

```
 return <Layout phenomicLoading={true} />
```

DavidWells:
I like `isLoading` but can see potential collisions with other peoples stuff. 

So perhaps `isPhenomicLoading` is better?

DavidWells:
http://www.screencast.com/t/8pqplVPSLxzQ

MoOx:
Very nice work :) That's definitely something we should put by default in the base theme.

Not sure about
- the direct access to ``__COLLECTION__``.
- the fact that we have 2 places where `<Layout` is used. We can probably just use the same place and just use the incomplete "page" object from the collection.

We will also need some tests + docs for this awesome feature!

DavidWells:
@MoOx Not sure how to get the data other than the `window. __COLLECTION__` but open to suggestions. This is probably ok for now until the collections refactor yes?

I can make docs + video for this no problem! 

I will need help on the tests. Not sure how to wire them up to check this.

MoOx:
Collection is accessible like in the normal app, via the context https://github.com/MoOx/phenomic/blob/bab74b8647ccfc20491f381303fdc340ae27ee24/src/client/index.js#L32-L37

Don't take time for video, I will work on integrating this into the default boilerplate and will setup an online demo for the base theme (see #424).

I can take care of tests if you want.

DavidWells:
Ok updated this to use collection off of context.

Can you do tests (if you want)?

Lets merge this baby!

MoOx:
Are you ok if I take it from here? Mainly to add tests, then maybe do some refactoring + doc.

DavidWells:
What needs to be refactored?

I'm not quite clear on where/how to test my changes in the test suite. Can you point me in the right direction?

MoOx:
First you should rebase on master to get latest changes on tests (ava > jest)
then you will need to add some tests here https://github.com/MoOx/phenomic/tree/master/src/components/PageContainer/__tests__ (probably just in components*)

I would like to find a way to merge https://github.com/MoOx/phenomic/pull/836/files#diff-4fbcb1b2984927ff93d3d2fcd07f9247R262 with line 274 (the other return <Layout)

We will also needs some docs here https://github.com/MoOx/phenomic/blob/master/docs/content/docs/usage/layouts.md

I also would like to make this the defaut way to display loading state in the base theme and want to provide 2 laoding component (the current loader + a text effect like on facebook - both should have the topbar).

MoOx:
Sorry if I am annoying but I think this small feature is really important so we should make this a featured feature :D

Brw we should also put some transitions examples on the base theme. This can be added after release :)

MoOx:
Working on a PR on this PR :)

MoOx:
You can follow my WIP here https://github.com/MoOx/phenomic/compare/master...new-loading (I didn't make a PR since I didn't rebase your current branch on master, but did on my "new-loading" branch).

I changed tons of stuff, mostly simple non important refactoring :)

PR #840, opened 2016-10-24
Add doc for internationalization

xuopled:


docs/content/docs/usage/intl.md
```diff
@@ -0,0 +1,236 @@
...
+const DEFAULT_LOCALE = "en"
+...
+let locale = window.localStorage.getItem("locale")
+
+if (locale !== null) {
+  store.dispatch(setLocale(locale))
+} else {
+  store.dispatch(setLocale(DEFAULT_LOCALE))
+}
+```
```
MoOx: If stuff like this, aren't you having different languages on a same url? So flash with default locale, then a rerendering with user locale?

xuopled: I forgot to add the doc about "redirect to home page".

But yes, maybe we can imagine to test two urls and redirect to the same url with user locale if the url exists

docs/content/docs/usage/intl.md
```diff
@@ -190,30 +213,35 @@ Once you can have your translated content, we have to add two buttons in the hea
...
   }
 }
+
+export default connect(
+  null,
+  dispatch => ({
+    ...
+    updateLocale: (locale) => {
+      dispatch(setLocale(locale))
+      browserHistory.push("/")
```
xuopled: @MoOx If we want to redirect to the same url with other locale, we have to do here

MoOx: I think it's preferable to do this, to avoid potential flash with different languages (eg: full page refresh will give you that (bad) UX). Thoughts?

docs/content/docs/usage/intl.md
xuopled: With this code, we have few seconds without translations. Do you think there is a way to add a general loader until the `SET_LOCALE` action is dispatched ?

MoOx:
Any plan to update the doc according to your discussion?

xuopled:
I will update the doc before the end of the week but I did not found a way to avoid the flash effect.
Maybe with #855, we could add a delay or handle the dispatch response.

xuopled:
I add a comment to explain why have uniq urls is a good thing. What do you think ?

MoOx:
What is the benefit of using redux for this? As you pass this to the context too.
I think we should better offer simpler instructions without redux (since we are going to remove it).

xuopled:
I don't know the benefit of using redux here.
I always used redux to manage my app state so I don't really know how to manage it without redux.
I need help to write the doc about it.

Maybe we can write 2 separate docs : 
  - with redux
  - without redux

Because I think a lot of people using react use also redux so they will be interested in it.

MoOx:
I am ok for

1. without redux
2.without redux

If you want, we can keep what you made for 2. and I can take care of 1. (I am going to do that for my personal website).

xuopled:
Ok I changed the label. Thank you for your help

MoOx:
Please next time use this repo instead of your fork, this will make contributions easier for me :)

xuopled:
Ok no problems :)

Issue #838, opened 2016-10-22
Linter errors in quickstart example


mdotasia:
I'm using v0.17.11 on Sierra and I'm running this example:

https://phenomic.io/docs/setup/

When I run `npm start` the site compiles fine and the browser opens as expected. But when I check my terminal I see tons of linter errors like these:

```
> phenomic start

ℹ Phenomic is starting +0.18s
✔ Development server listening on http://0.0.0.0:3333/ +0.72s
✖ Updated with warnings

⚠ Warning in ./scripts/phenomic.browser.js

./scripts/phenomic.browser.js
  2:1  error  Definition for rule 'react/no-find-dom-node' was not found          react/no-find-dom-node
  2:1  error  Definition for rule 'react/display-name' was not found              react/display-name
  2:1  error  Definition for rule 'react/jsx-no-undef' was not found              react/jsx-no-undef
  2:1  error  Definition for rule 'react/jsx-uses-react' was not found            react/jsx-uses-react

```

MoOx:
We found the issue and added a regression test so this won't happen in the future. Thanks for reporting this issue. Please follow #841.

Meanwhile, just replace this

```js
  "eslintConfig": {
    "extends": "./node_modules/phenomic/lib/eslint-config-recommended/index.js"
  },
```

by this


```js
  "eslintConfig": {
    "plugins": [ "react" ],
    "extends": "./node_modules/phenomic/lib/eslint-config-recommended/index.js"
  },
```

MoOx:
Closed by #841, 0.17.12 base theme don't have this issue anymore.

PR #841, opened 2016-10-24
Base theme linting should break the tests

MoOx:
Ref #838


MoOx:
The idea is to link the base theme with strict rules like the docs and core code but keep a less strict (standalone) config for new setup.

MoOx:
@montogeek Can you push your commit on this branch? Credit to you ;)

montogeek:
Sure!

Issue #845, opened 2016-10-26
Add more extensions in default webpack config for file loaders


MoOx:
https://gitter.im/MoOx/phenomic?at=5810cb1c482c168b22ba8f4c

PR #847, opened 2016-10-26
Added more extensions to Webpack

montogeek:
Fixes https://github.com/MoOx/phenomic/issues/845

We could also add extensions for Audio files: https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/config/webpack.config.dev.js#L161


MoOx:
Lgtm as it is. Thanks!

PR #849, opened 2016-10-27
Add new stylelint rule

xuopled:
`css-modules` are enabled by the default `webpack` configuration. 

 The `composes` rule is very usefull to compose selectors. I think it could be great to enable it directly as a `stylelint` rule

### DIFF
--- a/src/stylelint-config-recommended/index.js
+++ b/src/stylelint-config-recommended/index.js
@@ -1,0 +1,15 @@
+// @flow
+
+module.exports = {
+  "extends": "stylelint-config-standard",
+  "rules": {
+    "property-no-unknown": [
+      true,
+      {
+        "ignoreProperties": [
+          "composes",
+        ],
+      },
+    ],
+  },
+}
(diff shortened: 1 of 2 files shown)


MoOx:
Maybe we can do like for eslint config and create a shareable file? (easier for update for user)

xuopled:
For me, only the `composes` rules was missing to start a project with `css-modules`. 
Do you have propositions about some others rules ?

MoOx:
Nothing more for now.

xuopled:
Do you want I create a personal repo with  shareable file ?

MoOx:
We will handle this as a real package when splitting phenomic as multi packages with a monorepo. For now you can do something similar to what we do for eslint
- https://github.com/MoOx/phenomic/tree/7823ad28b726ddd2bd4161c262d1168be720dd4b/src/eslint-config-recommended
- https://github.com/MoOx/phenomic/blob/7823ad28b726ddd2bd4161c262d1168be720dd4b/themes/phenomic-theme-base/package.json#L85

MoOx:
Perfect! Will merge for next release, probably next week.

kevin940726:
This rule was removed in #876 , so the `composes` keyword error comes back. Should we fix it now? Perhaps with [stylelint-config-css-modules](https://github.com/pascalduez/stylelint-config-css-modules).

MoOx:
Default config is now pretty empty https://github.com/MoOx/phenomic/blob/master/src/stylelint-config-recommended/index.js. You should probably use your own config.

PR #854, opened 2016-10-30
DX improvement: Reuse tab if one is already open

DavidWells:
Snagged from https://github.com/facebookincubator/create-react-app/blob/0bd593baa777606ffb72f63eab9f5fd491062007/packages/react-dev-utils/openBrowser.js

This reuses the same browser tab if one is already open instead of reopening a new tab each time `phenomic start` runs.

For OSX only but it defaults back to `opn` package if person on linux/windows.

Lots of tabs getting opened was getting annoying and would sometimes cause funky stuff to happen with hot module reloading

### DIFF
--- a/src/builder/openChrome.applescript
+++ b/src/builder/openChrome.applescript
@@ -1,10 +1,5 @@
 (*
-Copyright (c) 2015-present, Facebook, Inc.
-All rights reserved.
-
-This source code is licensed under the BSD-style license found in the
--- LICENSE file in the root directory of this source tree. An additional grant
-of patent rights can be found in the PATENTS file in the same directory.
+Ripped from https://gist.github.com/mayoff/1138816
 *)
 
 on run argv
@@ -23,8 +18,9 @@ on run argv
       set theTabIndex to 0
       repeat with theTab in every tab of theWindow
         set theTabIndex to theTabIndex + 1
+        # found url from the same path, not only exact urls
+        # if theTab's URL is theURL then
         if theTab's URL contains theURL then
-        #if theTab's URL is theURL then
           set found to true
           exit repeat
         end if
--- a/src/builder/openChrome.applescript
+++ b/src/builder/openChrome.applescript
@@ -1,0 +1,49 @@
+(*
+Copyright (c) 2015-present, Facebook, Inc.
+All rights reserved.
+
+This source code is licensed under the BSD-style license found in the
+-- LICENSE file in the root directory of this source tree. An additional grant
+of patent rights can be found in the PATENTS file in the same directory.
+*)
+
+on run argv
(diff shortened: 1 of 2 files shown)


DavidWells:
This is step 1. =)

Step 2 would be reusing the same tab AND staying on the same path (`/some/url/here`) instead of taking people back to the homepage  `/` everytime

montogeek:
I was looking for this time long time!! :D

DavidWells:
Ok got it working and reusing the same tab with the same URL.

One difference here from the previous setup is that if the URL is already open in the browser `phenomic start` doesn't pop open the browser over your editor every time. It only focuses the browser the first time if the localhost:333x isn't open in chrome yet.

I personally like this better than having the browser focused each time I restart.

@MoOx thoughts on this?

I can probably make the applescript focus the browser no matter what (like the current setup) but I'm not sure we should =)

src/builder/openChrome.applescript
```diff
@@ -0,0 +1,49 @@
+(*
+Copyright (c) 2015-present, Facebook, Inc.
+All rights reserved.
+
+This source code is licensed under the BSD-style license found in the
+-- LICENSE file in the root directory of this source tree. An additional grant
+of patent rights can be found in the PATENTS file in the same directory.
+*)
```
MoOx: I don't want to be annoying, but I just don't want a explicit copyright in a file, that is already a copy of another (copyrighted, as there is not explicit license ?) script (https://gist.github.com/mayoff/1138816, 5yo gist).
Btw, it seems there is an improved version in a comment https://gist.github.com/mayoff/1138816#gistcomment-1821795 (not sure what the real diff)

src/builder/server.js
```diff
@@ -196,7 +197,33 @@ export default (config) => {
...
+          }
+          catch (err) {
+            // Ignore errors.
+          }
+        }
+        // Fallback to opn
+        // (It will always open new tab)
+        try {
+          opn(openUrl)
+          .catch(() => {}) // Prevent `unhandledRejection` error.
```
MoOx: Why is there a possible error?

DavidWells: Not sure. I saw it in the other setup and I'm sure it handles some error they saw

src/builder/openChrome.applescript
DavidWells: I saw probably 5 other implementations of the script and I'm not sure who was first to give attribution to.

At the end of the day, this code is just a system level utility so I'm not sure why you'd need copyrights on it.

Up to you what you want to do with it. All I know is that I want this to make my Dev flow faster :)

MoOx:
@DavidWells I pushed nano changed according to my comment to be able to ship this. Thanks again for you contribution <3.

Issue #587, opened 2016-08-01
Pages transitions


xuopled:
It could be great to handle pages transitions using css transitions to reproduce this kind of transitions : http://tympanus.net/Development/PageTransitions/

I used react-router to handle the routing transitions but there is maybe a better way : https://gist.github.com/cedricdelpoux/4286de789e7b0138708a3be4581963f2

What do you think ?

MoOx:
Will take a look when I got some times :)

cedricdelpoux:
Thank you for your help :)

DavidWells:
👍 looks cool!

MoOx:
I got an awesome idea:

- stop unmounting current Layout, instead keep it and just send new props (people will be able push an optimistic update with title, meta, with just the body to update, when data are received)
- handle render of the Loading differently (have to think about this a bit more)

This should allow people to do whatever they want (eg: use ReactCSSTransitionGroup) in they layout when props change.

cedricdelpoux:
Awesome solution. thank you for your help

MoOx:
I guess what we have in 0.18.0 is enough for now. We can improve and revisit later.

PR #855, opened 2016-10-31
Optimistic loading (new way to display loading state)

### LINKED ISSUE #824: UX improvement: Even FASTER Page Loading
Right now we have the [generic loading component](https://github.com/MoOx/phenomic/blob/36b03832dde5cf3a11d1b3480f15935172697060/src/components/PageContainer/component.js#L228) which is good but I was thinking about ways to make loading appear even faster for visitors.

Facebook loads the layout with phony loading state text. Example:

image

The effect makes the pages feel like they are loading faster to the user

**So the idea here is:**

If you build your layout components with default loading states, we could basically show the layout instantly when a link is clicked.
1. Visitor clicks link
2. Layout XYZ with prop `phenomicLoading={true}` (or something) renders instead of default generic `<PageLoading>` component
3. After `xyz.json` data is finished fetching & loads, the real layout component just refreshes with the content
4. Visitor is flipping amazed at how fast this site is

While this doesn't technically speed up anything, the perceived site performance would be much faster.

**What do you think?**

Possible implementation: 

We could check [here](https://github.com/MoOx/phenomic/blob/36b03832dde5cf3a11d1b3480f15935172697060/src/components/PageContainer/component.js#L228) is the requested Layout has the reserved `phenomicLoading` propType defined and if so, use it as loading component. If `phenomicLoading` propType is not defined use the default generic `<PageLoading>` component

B/c this would require some work/logic in layouts, this would of course be optional and not required if people just want a generic loading state component as it works now.

Related: https://github.com/MoOx/phenomic/issues/587

MoOx:
Now, if you remove your ``PageLoading`` layouts, you will automatically receive partial data in the correct (future) layout (only head, body won't be defined until ``isLoading`` is != true).

phenomic-optimistic-loading

Should not be a breaking change if you do still use ``PageLoading``.

If you still use ``PageLoading``, you can define a static property in some layouts to tell Phenomic to send an ``isLoading`` props to your specific layout instead of using ``PageLoading`` for this specific transition.

You can refer to [the minimal change you can do in the docs code](https://github.com/MoOx/phenomic/commit/66884e13a69ce6d30505454b763faca543c62b2c) if you want to replace ``PageLoading`` usage by optimistic loading via ``isLoading`` props.

This probably allows you to handle page transition (ref #587) since current layout won't be unmounted.

Closes #836
Closes #824


MoOx:
@DavidWells does this looks good to you?

DavidWells:
Looks great!

MoOx:
Next time feel free to approve the PR via the dedicated github feature :)

montogeek:
Oh la la!

DavidWells:
@moox Where is that button? =P I've haven't ever had to use that.

PR #867, opened 2016-11-07
Don't intercept clicks on links with no href

grncdr:
It's possible to have an anchor tag without an `href` attribute, and this code would have thrown an exception in that case.


MoOx:
Nice catch! Thanks for fixing this bug!

PR #898, opened 2016-11-29
Fix all the ci

MoOx:
- [x] remove yarn.lock files, as we miss semver failures on travis (we are not a end product, more a lib so we don't control deps - we just can't)
- [x] fix circle-ci mess with webpack 2
- [x] fix windows posinstall step
- [x] remove duplicate install step in scripts/* (only keep global postinstall)
- [x]  fix CI failure threshold to 70% (so add more tests yeah)


MoOx:
Coverage threshold is now up to 80% 😇

PR #903, opened 2016-12-02
Improve netlify-cms doc

### LINKED ISSUE #891: Netlify CMS instructions
I tried following the directions on the website for adding Netlify CMS, but ran into trouble.
https://phenomic.io/docs/usage/cms/netlify/

It says:
`Note: PHENOMIC_SOURCE & ASSETS_FOLDER are configurable in your package.json`

but I couldn't find those variables set anywhere. Ended up just hard-coding in content/assets .

That got the Netlify CMS to finally load, but when I tried creating a new post, it would just hang and nothing new would show up in my github repository.

Other things were unclear in the instructions. It says to add the following, but there were other things listed under plugins. Does order matter? I just put it at the bottom.
```
plugins: [
    new CopyWebpackPlugin([
      {from: 'admin', to: 'admin'},
    ]),
  ]
```

Under "Netlify configuration," for creating a "new developer application" I did that under "Developer settings" -> "OAuth applications." I think that's right, but it should probably be said explicitly. 

Thanks for all you do.

xuopled:
fix #891

### DIFF
--- a/docs/content/docs/usage/cms/netlify.md
+++ b/docs/content/docs/usage/cms/netlify.md
@@ -42,23 +42,34 @@ with 2 files :
   </html>
   ```
 
-2. `config.yml`:
-  _Note: PHENOMIC_SOURCE & ASSETS_FOLDER are configurable in your `package.json`.
-  Replace those in the section below_
+2. `package.json`
+
+  You need to configure the `source` and `assets` folders.
+
+  ```json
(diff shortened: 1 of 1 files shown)


MoOx:
Not sure it's more clear. What about directly using real paths ?

xuopled:
I updated

PR #904, opened 2016-12-02
Prevent duplicate loading of JS files

timothycoy:
When doing a build react-helmet is loading js files once in the head and once in the body.  React helmet manages the head and thus this script toComponent() should be in the head.

### DIFF
--- a/src/components/Html/__tests__/__snapshots__/aphrodite.js.snap
+++ b/src/components/Html/__tests__/__snapshots__/aphrodite.js.snap
@@ -23,6 +23,13 @@ exports[`test should render Html component with aphrodite styles when possible 1
     <link
       href="test.css"
       rel="stylesheet" />
+    <script
+      dangerouslySetInnerHTML={
+        Object {
+          "__html": "window._aphrodite = [\"stuff\"]);",
+        }
+      }
+      data-react-helmet={true} />
   </head>
   <body>
     <div
@@ -34,13 +41,6 @@ exports[`test should render Html component with aphrodite styles when possible 1
       id="phenomic" />
     <script
       phenomicStuff={true} />
-    <script
-      dangerouslySetInnerHTML={
-        Object {
-          "__html": "window._aphrodite = [\"stuff\"]);",
-        }
-      }
-      data-react-helmet={true} />
     <script
       src="test.js" />
   </body>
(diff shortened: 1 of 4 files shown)


MoOx:
Hi. Is this really creating an issue? Can you give more details about this?

timothycoy:
Yes, particularly because of the base head is not part of the template, so the best way to load js files in the head is to use react-helmet.  But then, because of this issue it's loading all the scripts twice.

Steps to reproduce:
npm run build
npm install --save-dev http-server
./node_modules/.bin/http-server dist

Screenshots:
screen shot 2016-12-03 at 10 02 55 am
screen shot 2016-12-03 at 10 03 10 am

MoOx:
Did you check if there are options in react-helmet about that?
I am not a fan on pushing by default script into the head as it's not a good practice (slow down the first paint)

timothycoy:
react-helmet is specifically for managing the head, so there are no options for that.  Also, sometimes you want to load scripts in the head.  For example, a script that has logic, which sometimes redirects to another page.  For example: optimizely.  Another example would be new relic, which you want to get accurate page load timings.

Now, with that said there are a number of other react libraries such as react-async-script or ReactScriptLoader that can be used to asynchronously load external scripts in the body.  But, these can easily be added as part of the template by the developer as needed.

arlair:
This change worked for me. My opinion is this commit sounds good as a temporary iterative step because currently the scripts are added to the head and the body. Then open a new enhancement issue to investigate body only script options.

I had a quick look and React-Helmet appears [hard coded to the headElement](https://github.com/nfl/react-helmet/blob/9a3fbdc063a6de3b68395767769f34e398fbe362/src/Helmet.js#L240) as far as I could tell. [headElement being set here](https://github.com/nfl/react-helmet/blob/9a3fbdc063a6de3b68395767769f34e398fbe362/src/Helmet.js#L240).

MoOx:
I am ok with this changes if it brings more problem than it's supposed to solves. Can you fix the relevant tests please? If you need some help [reach us on gitter](https://gitter.im/MoOx/phenomic)

timothycoy:
Done.

MoOx:
Thanks! I will release this as a minor (not a patch) as some people might get unexpected results.

PR #930, opened 2017-01-04
Create www.webstackdev.com.md

WebStackDev:


MoOx:
I remove "docs" (since your website is not a documentation website - correct me if I am wrong) and "code" since it's too generic (I would like to have a curated list only).

MoOx:
Nice website btw :)

Issue #933, opened 2017-01-09
Add meta for Twitter card images


VagishVela:
If you use the stock phenomic hero, since it is a background image it doesn't get picked up by twitter (and probably Facebook as well).

Some complexity would be figuring out whether the URL is a local path or on another website.

MoOx:
User land thing.

PR #934, opened 2017-01-09
Add twitter card image support

VagishVela:
So far started working on it, just creating a PR to track status of #933.

- [x] Add support for local paths
- [x] Add support for non-local paths
- [x] Account for protocol-less urls
- [ ] Move regexp test to the loader
- [ ] Appropriate tests

### DIFF
--- a/themes/phenomic-theme-base/src/layouts/Page/index.js
+++ b/themes/phenomic-theme-base/src/layouts/Page/index.js
@@ -42,6 +42,7 @@ const Page = (
     { name: "twitter:title", content: metaTitle },
     { name: "twitter:creator", content: `@${ pkg.twitter }` },
     { name: "twitter:description", content: head.description },
+    { name: "twitter:image", content: joinUri(process.env.PHENOMIC_USER_URL, head.hero) },
     { name: "description", content: head.description },
   ]
 
--- a/themes/phenomic-theme-base/src/layouts/Page/index.js
+++ b/themes/phenomic-theme-base/src/layouts/Page/index.js
@@ -30,6 +30,10 @@ const Page = (
 
   const metaTitle = head.metaTitle ? head.metaTitle : head.title
 
+  const externalUrlCheck = new RegExp('^https?://');
+  const socialImage = externalUrlCheck.test(head.hero) ?
+  head.hero : joinUri(process.env.PHENOMIC_USER_URL, head.hero)
+
   const meta = [
     { property: "og:type", content: "article" },
     { property: "og:title", content: metaTitle },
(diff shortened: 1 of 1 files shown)


MoOx:
Nice start!

Could you add as well og support?

VagishVela:
@MoOx Sure thing.

themes/phenomic-theme-base/src/layouts/Page/index.js
```diff
@@ -30,22 +30,24 @@ const Page = (
 
   const metaTitle = head.metaTitle ? head.metaTitle : head.title
 
+  const externalUrlCheck = new RegExp('^https?://');
+  const socialImage = externalUrlCheck.test(head.hero) ?
+  head.hero : joinUri(process.env.PHENOMIC_USER_URL, head.hero)
+
```
thangngoc89: I think we  should move this into the loader part, processing this on every render isn't good for performance wise

thangngoc89: Also, i usually use protocol-less url (//example.com/img.jpg)

thangngoc89:
Also, the process.env.PHENOMIC_USER_URL is always what you defined in `package.json` homepage url. Please handle the development mode.

MoOx:
We should probably handle that. The theme should not have code for development imo.

thangngoc89:
So PHENOMIC_URL. I guess? But the port can't be known before starting the dev server

MoOx:
We should not move anything to the loader as it's going to disappear.

VagishVela:
@MoOx @thangngoc89 So where should I handle https://github.com/VagishVela/phenomic/blob/261c9cc08e8c4d3b6f38f20e296171ce709e5b0b/themes/phenomic-theme-base/src/layouts/Page/index.js#L33-L35 ?

themes/phenomic-theme-base/src/layouts/Page/index.js
```diff
@@ -30,18 +30,24 @@ const Page = (
 
   const metaTitle = head.metaTitle ? head.metaTitle : head.title
 
+  const externalUrlCheck = new RegExp('^(https?:)?\/\/');
+  const socialImage = externalUrlCheck.test(head.hero) ?
+  head.hero : joinUri(process.env.PHENOMIC_USER_URL, head.hero)
```
MoOx: Could you simplify to

```js
const socialImage = head.hero.match("://") ? head.hero
  : joinUri(process.env.PHENOMIC_USER_URL, head.hero)
```

MoOx:
Thanks for this contribution!

PR #340, opened 2016-03-22
Use node-portfinder to avoid error when port is used

thangngoc89:
Close #320

Issue #888, opened 2016-11-22
Create a light phenomic-cli package to ease setup


MoOx:
Will be handled in #598.

We need to create a small cli to make setup more easy. The package must be super light and should not have lot's of logic to avoid updates (even if update notifiers exist)

bloodyowl:
Going to be fixed with #925

MoOx:
Closing in favour of #734

Issue #734, opened 2016-09-06
phenomic-cli: make setup fast & more easy, (one install, one command?)


MoOx:
We need to improve that, current setup page is not appealing.

Maybe create a phenomic-cli or something when handling #598.

Something fast, so not a electron app yet :D

thangngoc89:
Which kind of work flow do you expect? 

npm install -g phenomic-cli
phenomic-cli new my-site

DanielFGray:
a yeoman generator could be a good choice

thangngoc89:
@DanielFGray did you try settting up a new phenomic project? We have everything in place, just need a better flow for new user

DanielFGray:
Yeah I've just started with phenomic. I really dig it, but the whole `DIR=your-website-folder && mkdir $DIR && cd $DIR && mkdir node_modules && npm i phenomic && ./node_modules/.bin/phenomic setup && npm i` is a bit much. yeoman should be able to handle all of that. it uses inquirer as well, which seemed to be in use for the setup script.

I have yet to make a yeoman generator myself, but I'll have a look into what it takes and see if I can't create one

thangngoc89:
@DanielFGray Thanks for your feedback. As you said the whole command sucks. 

My idea so far is creating a global phenomic cli that should never changed. If user want to create a new phenomic project, this is what they have to do: 

```
$ npm i -g phenomic-cli
$ phenomic new folder-name
```

This will run `DIR=your-website-folder && mkdir $DIR && cd $DIR && mkdir node_modules && npm i phenomic && ./node_modules/.bin/phenomic setup && npm i` 


Sounds good?

DanielFGray:
It doesn't sound bad in the slightest, I just can't help thinking there's a certain amount of re-inventing the wheel there, and that off-loading it to yeoman could be a lot less work than writing a `phenomic-cli` project from scratch.

Just my two cents, I would be happy to use either.

MoOx:
First keep in mind that the first part of the command (`DIR=your-website-folder && mkdir $DIR && cd $DIR && mkdir node_modules`) is the kind of instructions that no project includes.
Most of the time, when you start a project, you create a folder by yourself, and sometimes, already have a package.json.
[...]

Issue #732, opened 2016-09-06
Generate routes for data coming in from outside phenomic


ben-eb:
Currently, pages such as http://cssnano.co/optimisations/discardDuplicates will respond with a 404 error code on initial load, then the routing will kick in and display the page. This is because the data for this page comes from outside phenomic:

https://github.com/ben-eb/cssnano/blob/2ad9e47263efc6581dec1cd45ad1b2e20b9d49ab/docs/web_modules/app/metadata.js#L27

The page is then routed here:

https://github.com/ben-eb/cssnano/blob/2ad9e47263efc6581dec1cd45ad1b2e20b9d49ab/docs/web_modules/app/routes.js#L32-L35

And the page itself:

https://github.com/ben-eb/cssnano/blob/2ad9e47263efc6581dec1cd45ad1b2e20b9d49ab/docs/web_modules/layouts/Optimisations/show.js#L18

Ideally, we should be able to tell phenomic that this is a real page somehow, so that a 404 isn't thrown.

MoOx:
I will implement a workaround so you can have this before we handle the big change related to #713 :)

bloodyowl:
Going to be fixed with #925

MoOx:
Coming in next release of v1

PR #936, opened 2017-01-12
1.0

### LINKED ISSUE #40: Consider replacing redux in favor of react context

### LINKED ISSUE #358: Generate page from json directly
Can i use it for display pages from json directly, not from markdown?
Use case: I get json that need statinamic from wordpress (through wp-api, and some modifications). Feed this json to statinamic (route setups manually or dynamic), and statinamic render my pages.
It's really to do?

### LINKED ISSUE #598: Explode as a monorepo

### LINKED ISSUE #609: Improve collection loading for huge websites
Just tested out a 1000 page site. Generated them with this handy script: https://gist.github.com/DavidWells/a75045c89d6f3d453ef4ab68499550f4

Good news, the build only took like 60 seconds! 

The generated site was 296mb in total though. This was primarily due to every page inlining the entire  `window.__COLLECTION__`  into the html. (https://gist.github.com/DavidWells/675b1b838c44925d6487fff3cc129674)

The `window.__COLLECTION__` makes html files weight 300k vs 8k

Perhaps there is a way to lazy load in the routing data and stick it inside localstorage or a service worker?

Or build the `window.__COLLECTION__` into a single `collection.js` file and include that once in the head tag? Then it could be cached and would make all the html files lighter.

### LINKED ISSUE #630: Plugins, round 2
In current master (will be released soon) we have ability to use plugins for phenomic-loader. Let's call that "phenomic loader plugins".
We will end up having theme (#508), so having a global plugin system seems like a good next step.

Here is my idea about what a plugin should be able to do:
- adjust config
- adjust webpack config (add loaders, plugins etc)
- register phenomic loader plugins

Please share your ideas!

### LINKED ISSUE #657: Pagination
The implementation will be a surprise ;)

### LINKED ISSUE #674: Ability to use multiple phenomic loader instances
This would allow you to have a mixed document style in your project (like markdown with ascii/ latex or whatever)

We will define a `useConfigKey` to determine where to load the config from webpack's config

### LINKED ISSUE #676: Automatic registeration of page layouts
Easy!

``` js
import React, { Component } from "react"
import { Route } from "react-router"

import AppContainer from "../components/AppContainer"
import PhenomicPageContainer from "phenomic/lib/PageContainer"

/* ! HERE IS THE TRICK */
const reqLayouts = require.context("../layouts", true, /\/index\.js$/)
const layouts = reqLayouts.keys().reduce((layouts, layout) => ({
  ...layouts,
  [layout.replace("./", "").replace("/index.js", "")]:
    reqLayouts(layout).default,
}), {})
/* ! */

class PageContainer extends Component {
  render() {
    const { props } = this
    return (
      <PhenomicPageContainer
        { ...props }
        layouts={ layouts }
      />
    )
  }
}

export default (
  <Route component={ AppContainer }>
    <Route path="*" component={ PageContainer } />
  </Route>
)
```

We should add this in the default theme and also try to reduce this code, maybe using a function that accept the "reqLayouts". An issue might be the context that will be used for the require (same thing as the hot reload for markdown...)

### LINKED ISSUE #700: Discussion: Support for YYYY-MM-DD-post-name.md markdown files
Would it be possible to support the standard jekyll post name format `YYYY-MM-DD-post-name.md`?
1. People could port from jekyll to phenomic with relative ease =)
2. Files in /blog would be easier to navigate through (example: https://github.com/auth0/blog/tree/master/_posts everything is in date order to quickly jump to a given post)

We are doing the serverless blog in phenomic as well and I'm trying to plan for the future when we have like 100+ markdown posts in a folder, It would be nice to have them in date order.

**Idea:** When phenomic builds routes, any markdown file name matching the pattern `YYYY-MM-DD` would automatically strip out the date and just be the name of file.

**Example:**  `2012-01-27-hello-world.md` become `hello-world` without needing to specify the custom `route` in the yaml frontmatter

Currently, you would need to provide the `route` param to frontmatter to the file with the `YYYY-MM-DD` prefix to remove that on the site url. I think we can streamline/remove this =)

### LINKED ISSUE #713: Improve the way we consume collection(s) to avoid sending all metadata in all pages
Related to trimming down the `window.collection`.  #712 

When you add in new custom meta (like in this blog post example) all those extra fields are also added to the `window.collection` data. This could potentially make it HUGE

Example:

`````` yaml

---
layout: Post
title: 'Defining Serverless and Why It Matters to Developers'
date: 2016-09-01
description: "You’ve probably heard the term _serverless._ But what does it actually mean? And more importantly, as a developer, why should you care?"
author:
  name: Serverless
  url: http://twitter.com/goServerless
  avatar: https://avatars3.githubusercontent.com/u/13742415?v=3&s=60
tags:
- serverless

---

```json
// printed in window.collection, notice the custom meta values
{
    "layout": "Post",
    "comments": true,
    "title": "Defining Serverless and Why It Matters to Developers",
    "date": "2016-09-01T00:00:00.000Z",
    "description": "You’ve probably heard the term _serverless._ But what does it actually mean? And more importantly, as a developer, why should you care?",
    "author": {
        "name": "Serverless",
        "url": "http://twitter.com/goServerless",
        "avatar": "https://avatars3.githubusercontent.com/u/13742415?v=3&s=60"
    },
    "tags": ["serverless"],
    "__filename": "blog/defining-serverless-and-why-it-matters-to-developers.md",
    "__url": "/blog/defining-serverless-and-why-it-matters-to-developers/",
    "__resourceUrl": "/blog/defining-serverless-and-why-it-matters-to-developers/index.html",
    "__dataUrl": "/blog/defining-serverless-and-why-it-matters-to-developers/index.html.bf1d5e0db467ef721b8508992749379b.json"
}
``````

All the data also exists in the `index.html.bf1d5e0db467ef721b8508992749379b.json` file. 

So the question is, **would it be possible/worth it** to remove the additional custom fields from outputting into `window.collection` and just have the .json files for the individual pages handle that additional data?

### LINKED ISSUE #778: Props not passed from PageContainer to specified layout
PageContainer includes a wealth of routing props such as params, history, etc.

I'd like to be able to grab some of them and pass it down to the layout that is actually rendering such as Homepage or Post.

I think this has to be handled by passing props to the createElement call on line 232 of component.js in the PageContainer lib module.

```
!!page.error && PageError && _react2.default.createElement(PageError, page),
        !page.error && !page.loading &&
        Layout && _react2.default.createElement(
           Layout,
          _extends({
             ref: this.saveContentRef },
             page,
             props)) // <= Adding in props here 
```

### LINKED ISSUE #780: Be able to consume data files
It would be great to create a folder of json/yml files and have them turned into a collection object so that we could arbitrarily loop through the data. 

This may be out of scope for this project but I thought as Jekyll and other static site generators it might be useful to have.

### LINKED ISSUE #791: Add support for multiple collection
I'd like to be able to define additional directories that get processed into their own collection. Like a gallery folder or something of that nature instead of depending on tags to filter out post types. Like a wordpress taxonomy.

### LINKED ISSUE #806: FAQ Consume more than one folder for the content
New api should offer a better way to do this, but meanwhile, you can use 2 phenomic loader instance with 2 differents context

in `webpack.config.babel.js`

``` js
{
  test: /\.md$/,
  loader: phenomicLoader,
  include: [
    path.resolve(__dirname, config.source),
  ],
  query: {
    context: path.join(__dirname, config.source),
  },
},
{
  test: /\.md$/,
  loader: phenomicLoader,
  include: [
    path.resolve(__dirname, "otherfolder"),
  ],
  query: {
    context: path.resolve(__dirname, "otherfolder")
  },
},
```

in `scripts/phenomic.browser.js`

``` js
// ...

let mdContext = require.context("../content", true, /\.md$/)
mdContext.keys().forEach(mdContext)
let mdContext2 = require.context("../otherfolder", true, /\.md$/)
mdContext2.keys().forEach(mdContext)

// also update code below for proper hot loading on markdown files
```

### LINKED ISSUE #889: New data/collection API
Currently our api is mostly a legacy from the initial version and is not really simple and pretty even if "it works". We have some limitations and some stuff are not optimal

- we have a single collection (that's not a big deal for me, since you can easily filter but I understand this can looks like a limitation) #791
- the entire collection data is in all HTML pages (we could move into a file and load it before starting the client side rendering, but again it's still not optimal ): we need to have a way to split in HTML only what is used #609 #713 #712
- it's not very easy to consume json files or external API (we could improve loader for json consumption but we will still be limited somehow) #806 #358 #780 #719 #886 #880
- we publicly rely on react context, which is not good and intuitive (lot's of people say "it's not working" and they just forget to  add contextTypes. We should hide this behind a provider component #304

I also want to add this

- we publicly use redux (some people are happy, but pretty sure most people don't care). We should hide what we use somehow #40
- we render content to html string which have some limitation, we should support more output, LIKE REACT COMPONENTS FOR EXAMPLE #434 
- routes should be resolved using not just markdown front-matter data #732

Currently we could use more tricks but we discussed already about smarter ideas on how to have a better intuitive, minimalist API which will even unlock feature like pagination #657

⚠️ I didn't mention graphql since I don't think it will be really intuitive for most people. Gastby 1.0 branch is probably doing awesome stuff with that, but I have to admit I found that a bit overkill, especially for simple websites or documentation. It's probably very powerful but maybe a bit too much for most cases when you have data in collection(s) and just want to iterate on it (blog, photo gallery etc). You tell me.

---

Fresh API ideas have been exposed here https://github.com/MoOx/phenomic/issues/713#issuecomment-244526898

I will start to work on something based on what @bloodyowl developed in this thread but feel free to comment and dissuss here about the approach YOU think is the best. There is no bad ideas, just ideas we can learn from ;)

MoOx:
This branch is contains current 0.x AND 1.0 (in `packages/*`)

---

If you want to play with alpha version (almost stable architecture architecture - but undocumented yet) you can open NEXT.md and read instructions, it's pretty easy thanks to lerna.

We will iterate here by doing PR on this PR until we are happy. We might eventually push this PR into master sooner and cut a "stable" branch to keep 0.x (for bugfixes only) until this 1.x is out (at least as a beta version).

---

closes #40 
closes #340 
closes #358 
closes #598
closes #609 
closes #630 
closes #657 
closes #674
closes #713 
closes #780 
closes #791 
closes #806 
closes #889 
closes #700

This will be closes because not relevant anymore
 
closes #676
closes #778 

Also open the door to #517 #732 #734 #888


lerna-debug.log
```diff
@@ -0,0 +1,52 @@
+lerna(verbose) GitUtilities.isInitialized    ()
```
thangngoc89: Could you remove this file? Useless diff

MoOx: Obviously it's too early to make an actual code review, lot's of stuff are not clean :)
Thanks for reporting.

MoOx:
What do you think about using npm namespace?

`@phenomic/core`, `@phenomic/plugin-react`, `@phenomic/plugin-*`... Will be easier to give npm rights access to anyone (just add to `@phenomic`). And seems legit. Thoughts?

MoOx:
We could keep "phenomic" for a version with good defaults and no config.

thangngoc89:
Yes. Using @phenomic is a very good idea. :D

MoOx:
Seems it's 14$ per month, can't find a free link for oss right now... I won't pay 170$/y for this for an open source project.

ben-eb:
`@phenomic/core` vs `phenomic-core` - they're not really that different aside from the price. 😄

MoOx:
The pain with no scope is when a contributor is added, you need to add npm rights for all packages... And when you create a new package, you need to think to add rights to all members etc. That's the main diff imo. Also it looks "official" (and that's another great thing).

bloodyowl:
@MoOx let's just write a script that reads a `OWNERS` and runs the command through `lerna exec`, don't you think? has the advantage of being declarative

MoOx:
Yeah, I think that's what babel does.

packages/phenomic-docs/App.js
```diff
@@ -0,0 +1,42 @@
...
+import ShowcaseList from "./components/ShowcaseList"
+import APITagListPage from "./components/APITagListPage"
+
+import { Renderer } from "phenomic-react/lib/client"
+
+const app = Renderer.createApp(
+  <Router history={ browserHistory }>
+    <Route component={ (props) => <div><Header />{props.children}</div> }>
+      <Route path="/" component={ Home } />
+      <Route component={ props => <div>{props.children}</div> }>
```
MoOx: @bloodyowl can you confirm this wrapper is useless here and for /changelog routes below?

bloodyowl: Yep

packages/phenomic-plugin-api-related-content/src/index.js
```diff
@@ -0,0 +1,25 @@
...
+    type: "api",
+    define(api) {
+      api.get("/related/:collection/limit-:limit/*.json", async function(req, res) {
+        try {
+          const limit = parseInt(res.params.limit)
+          const post = await req.db.get([ req.params.collection ], req.params[0])
+          const lists = await Promise.all([
+            ...post.value.tags.map(tag => req.db.getList(req.params.collection, { limit: limit + 1, lt, reverse }, "tags", tag)),
+            db.getList(req.params.collection,  { limit: limit + 1 }),
+          ])
```
MoOx: @bloodyowl can you tell me what lt and reverse are supposed to be/come from?
Also, below I guess it's req.db.getList, not just db.getList? (I am enabling the linter so I am spotting this kind of issue - which allow me to be familiar with the code)

bloodyowl: yeah, should be `db`.

- `lt` -> less than
- `reverse` -> reverses the query order

they both come from https://github.com/Level/levelup

vvo:
I would not go for scoped packages, from recent experience it seems that ultimately not every tooling is ready to read package.json fine with scoped packages (especially yarn)

MoOx:
I already subscribed for one ;)

packages/phenomic-plugin-renderer-react/src/render/server.js
```diff
@@ -0,0 +1,63 @@
...
+      error
+      ? reject(error)
+      : resolve({ renderProps, redirectLocation })
+    })
+  })
+}
+
+function renderToString(store, { renderProps, redirectLocation }, renderHTML) {
+  const body = ReactDOMServer.renderToString(
+    <Provider fetch={ fetch } store={ store }>
```
MoOx: @bloodyowl can you explain me here how this fetch ("spec" fetch) can do his job here when Provider except another kind of fetch (the one that return .json()). I am probably too tired to understand this (I am finishing to add flow types everywhere in the codebase and this is the only issue left for me :) )

packages/phenomic-plugin-transform-markdown/src/remark-plugins.js
```diff
@@ -0,0 +1,50 @@
...
+    .use(autoLinkHeadings, {
+      // @todo find how to make this options works with remark-react
+      content: {
+        type: "text",
+        value: "#",
+      },
+      linkProperties: {
+        className: "phenomic-HeadingAnchor",
+      },
+    })
```
MoOx: poke @ben-eb :)

packages/phenomic-plugin-transform-markdown/src/__tests__/__snapshots__/remark-plugins.js.snap
```diff
@@ -0,0 +1,57 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`should render markdown as react component 1`] = `
+<div>
+  <h2
+    id="test"
+  >
+    <a
+      href="#test"
+    />
```
MoOx: and poke again @ben-eb

MoOx:
Hey everyone, you can expect a first alpha at the end of the month! I am sure @revolunet will be happy :D

revolunet:
haha no pressure guys :) but YES i'm very happy thats a great news thank you 👍

MoOx:
Bye v0.x, we are going to focus on v1 in master! (tests are broken but no big deal, we will fix this asap)

PR #938, opened 2017-01-14
ftypo

Maxhodges:
the path is incorrect. needs the `.html` extension to work.


MoOx:
It has been incorrectly updated during 79c4058025d88c352dace750363da2d80f613b8c.

Thanks for catching this and for the PR ❤️

PR #946, opened 2017-01-20
Update offline-plugin to the latest

### LINKED ISSUE #905: offline-plugin v4 release
@MoOx hey! I recently released `v4` of `offline-plugin`. Would you like to update to it or you are comfortable with you custom branch?

NekR:
Closes #905

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "mkdirp": "^0.5.1",
     "multili": "^1.0.1",
     "node-object-hash": "^1.0.2",
-    "offline-plugin": "https://github.com/MoOx/offline-plugin#phenomic",
+    "offline-plugin": "^4.5.5",
     "opn": "^4.0.2",
     "ora": "^0.3.0",
     "pify": "^2.3.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1876,10 +1876,6 @@ es6-map@^0.1.3:
     es6-symbol "~3.1.0"
     event-emitter "~0.3.4"
 
-es6-promise@^3.0.2:
-  version "3.3.1"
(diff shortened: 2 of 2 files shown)


MoOx:
Thank you!

MoOx:
Actually it seems with latest version of something, it's not working on node 4. Any idea @NekR ? https://travis-ci.org/MoOx/phenomic/jobs/195448787#L1010

NekR:
Hmm.. Not sure, I have all tests passed on Node 4 both on Windows and Linux. Might it be something else, not offline-plugin?

MoOx:
We are using yarn so the only diff for the passing and not passing build should be the version of node.
https://travis-ci.org/MoOx/phenomic/builds/195448785
Here passing for node 6 and not for node 4.

NekR:
Yeah, I see it throws `Error: OfflinePlugin: Plugin's runtime wasn't added to one of your bundle entries. See this https://goo.gl/YwewYp for details.`, that's weird especially that it happens on Node 4 only. I'll debug it.

MoOx:
Thanks for your help.

NekR:
@MoOx btw, can you try re-build on that test? Because I don't see it failing on my commit but failing on `Base theme: Add fb/twitter card image support` commit so that's totally weird.

NekR:
Okay, just tried it, but I'm not able to repro this locally on macOS with Node 4. Are you able to do so?

MoOx:
I am currently able to reproduce this.

MoOx:
I found the issue. In short it's because of the way phenomic is using itself (there is 2 offline-plugin in 2 different node_modules and that's causing an issue). Weird it's "random". Thanks for your time and sorry for the false alarm!

PR #951, opened 2017-01-27
0.20.3

MoOx:


MoOx:
I just miss webpack 2.2 stable support and this PR should be good.

PR #961, opened 2017-02-06
Update context to users on the homepage

bdougie:
Just noticed this, perhaps it should be users?


bdougie:
I also removed the extra proposition

MoOx:
Thanks!

PR #962, opened 2017-02-07
Wrap setup shell command variables in quotes

lytedev:
In the setup, it is advised to store the desired site directory in a variable for later use, then the variable is used later for setting up paths. This would break the command if the path contains any spaces and is generally considered bad practice. Quoting your shell variables when using them (especially when paths are concerned) is [highly][ref1][[1][ref1]] [recommended][ref2][[2][ref2]].

I can't remember if Windows acts the same.

\[1]: http://www.tldp.org/LDP/abs/html/quotingvar.html
\[2]: http://stackoverflow.com/questions/10067266/when-to-wrap-quotes-around-a-variable

[ref1]: http://www.tldp.org/LDP/abs/html/quotingvar.html
[ref2]: http://stackoverflow.com/questions/10067266/when-to-wrap-quotes-around-a-variable

### DIFF
--- a/docs/content/docs/setup.md
+++ b/docs/content/docs/setup.md
@@ -26,13 +26,13 @@ we use html just to add contenteditable
 
 ```sh
 DIR=your-website-folder
-mkdir $DIR && cd $DIR && mkdir node_modules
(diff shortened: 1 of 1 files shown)


MoOx:
Thanks!

PR #970, opened 2017-02-20
shorten shell commands in doc

doodzik:
shorten shell commands in the docu by using the $_ variable.

http://unix.derkeiler.com/Newsgroups/comp.unix.shell/2008-10/msg00283.html

### DIFF
--- a/docs/content/docs/setup.md
+++ b/docs/content/docs/setup.md
@@ -25,8 +25,7 @@ _Tip: you can edit ``your-website-folder`` in the text box below_.
 we use html just to add contenteditable
 
 ```sh
-DIR=your-website-folder
-mkdir $DIR && cd $DIR && mkdir node_modules
+mkdir "your-website-folder" && cd "$_" && mkdir node_modules
 ```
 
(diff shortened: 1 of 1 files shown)


MoOx:
Does this work with zsh?

doodzik:
@MoOx yes

MoOx:
Thanks (I learned a thing) but I will pass. I am afraid it add more confusion. Anyway, 1.0 will have a tool to avoid this long command ;)

doodzik:
What is your estimate on the release date of 1.0?

What do you think of stripping the docu from fill words (just, simple, easy, ect.) to make it sound more technical?

MoOx:
@doodzik no ETA for 1.0 yet.

I don't want a 100% technical doc, but I guess we can do the following:

Getting Started: noob friendly not 100% technical + Technical docs ?

doodzik:
Ok. We can evaluate it for a page in the docs and evaluate it then.
But I would do it after the released, because the docs are going to be change.

Issue #973, opened 2017-02-25
HMR error when editing markdown files requires a full reload after the first change


evenkiel:
HMR live reloads stop working after the first change to an *.md file. To reproduce in the base default project/theme, edit the index.md file. Note that after the first change (which does live reload), you'll see the following error output in the chrome console:

<img width="664" alt="screenshot_29" src="https://cloud.githubusercontent.com/assets/1083590/23334459/87af132c-fb64-11e6-9542-437a5fb85892.png">

Subsequent changes to the same file will not cause the browser to live update.

I've identified the root cause and have a fix locally which resolves. Unfortunately I'm having trouble building from source on windows and keep getting a post-install error. Once I'm able to push past the build issue I plan on submitting a PR.

I'll cross post on Gitter chat in case someone there knows how to resolve the build problem.

MoOx:
Seems to be a dup of #779

PR #974, opened 2017-02-25
Fix HRM hot reload for md files (#973)

evenkiel:
When making changes to md files I noticed that an HMR error would throw after the first change, and thereafter you would have to force reload the page in the browser to see updates. Added simple null check down in hot-md to prevent the error from throwing.

Tested locally and verified that hot reloads worked properly with md files.

### DIFF
--- a/src/client/hot-md.js
+++ b/src/client/hot-md.js
@@ -6,6 +6,9 @@ export default (mdContext, collection, store) => (file) => {
     (item) => item.__filename === file.slice("./".length)
   )
   const dataUrl = mdContext(file)
+  if (null == item){
+    return;
+  }
   if (dataUrl !== item.__dataUrl) {
     item.__dataUrl = dataUrl
     console.log(file, " hot update")


mktange:
I also just encountered this issue. However, I don't think this PR solves the issue of hot-reloading markdown files on Windows - it just removes the HMR error.

The issue is that the `__filename` variable in the collection looks like `path\to\file.md` on Windows, while the file paths gotten from the webpack `require.context` are of the form `path/to/file.md`.

A solution that is working for me has been to normalize the `__filename` paths of the collection in the phenomic loader. This way the hot-md reload is able to find the matching file:

https://github.com/MoOx/phenomic/blob/master/src/loader/index.js#L68

changed to

```javascript
...
  const metadata = {
    __filename: pathToUri(relativePath),
    __url: pathToUri("/", url),
...
```

MoOx:
Closed by #1005

PR #994, opened 2017-04-01
Path to phenomic fix for Windows

mbeaudru:
I might be wrong, but as I tried to do the setup on a Windows 10 machine I found out that the current paths aren't recognized by windows. I did the change I suggest and everything worked fine.


MoOx:
cmd? powershell?

mbeaudru:
I tried with cmd & powershell and the current path does work, I was using Git Bash and that's why it was not working, sorry.

As many devs are using Cmder / Git Bash on Windows, maybe we should add something about running this command with cmd / powershell in the instructions ?

I close this PR as this is not the right solution, but if you agree with the above statement I can submit a PR 👍

MoOx:
> As many devs are using Cmder / Git Bash on Windows, maybe we should add something about running this command with cmd / powershell in the instructions ?

Yes please! :)

Issue #944, opened 2017-01-19
Yaml Frontmatter Date off by 1 day


DavidWells:
Just noticed this on the serverless.com blog. Looks like the date from `head` data is shifting the date one day backwards.

This frontmatter:
```
---
date: 2016-12-15
---
```

In the Post layout `head.date` === `2016-12-15T00:00:00.000Z`

When converting this to date `new Date(head.date)` it gives back `Wed Dec 14 2016 16:00:00 GMT-0800 (PST)` which is a day off

Is anyone else seeing this behavior?

This is the code printing out our date: https://github.com/serverless/site/blob/a1659dafc8ceffd3e5df5b14086eb63cffb90635/src/layouts/Post/index.js#L37-L46

I can fix this on my end with some math but it seems like a bug that might be effecting others.

This will shift the date to the correct `2016-12-15` date
```
console.log(new Date(pageDate.getTime() + Math.abs(pageDate.getTimezoneOffset() * 60000)))
```

Here is another example: https://serverless.com/blog/a-serverless-windclock-at-the-lake/ and the markdown https://github.com/serverless/blog/blob/master/posts/2017-01-17-a-serverless-windclock-at-the-lake.md (the date should be 2017-01-17 but it's not =P)

DavidWells:
Live example links won't work. I fixed this in userland: https://github.com/serverless/site/commit/0a9bc4045b63997f7fef17ae4e96b53d414f344b but still think this might be a core issue

MoOx:
How can it be a core issue? Date are strings, so it's impossible Phenomic can be responsible of this.
If I run this in my browser console

```
> const d = new Date("2016-12-15")
> d.toISOString()
< "2016-12-15T00:00:00.000Z"
```
I get a correct number.

I don't have this issue on others projects that use the same technique.

Even the default theme is not affected.

- https://github.com/MoOx/phenomic/blob/89ce2f8d803469a47bb450fd6e0b4f4d6d88c001/themes/phenomic-theme-base/content/posts/hello-world.md
- https://phenomic.io/themes/base/demo/posts/hello-world/

<img width="695" alt="screen shot 2017-01-19 at 07 00 33" src="https://cloud.githubusercontent.com/assets/157534/22095457/13d9cbfc-de15-11e6-81a6-688e9f0c5407.png">
[...]

PR #999, opened 2017-04-03
Base theme: use date-fns to parse post dates

kristerkari:
## Why ?

Currently `new Date` is used to format dates for base theme posts. This works when the post date is in `yyyy-mm-dd` (e.g. `2016-01-22`) format, but for example Safari will throw an error if you try to add hours and minutes (without seconds, e.g. `2016-01-22 14:31`) and seems to also return `Invalid Date` for many of the combinations that I have tried. Chrome seems to be a bit more relaxed with the dates, but will also return `Invalid Date` for some date combinations.
 
I think that since it is not obvious which formatting style throws and which doesn't, it would be good to try to support [the formats that are mentioned in this blog post](http://www.pelagodesign.com/blog/2009/05/20/iso-8601-date-validation-that-doesnt-suck/). 

## How does using date-fns change the situation?

Date-fns' parse also accepts dates that are not in the full ISO 8601 format:

```
Convert the given argument to an instance of Date.

If the argument is an instance of Date, the function returns its clone.

If the argument is a number, it is treated as a timestamp.

If an argument is a string, the function tries to parse it. 
Function accepts complete ISO 8601 formats as well as partial implementations. ISO 8601: http://en.wikipedia.org/wiki/ISO_8601

If all above fails, the function passes the given argument to Date constructor.
```
https://github.com/date-fns/date-fns/blob/master/src/parse/index.js

## Change in file size

Before:
```
-rw-r--r--   1 krister  staff   298K Apr  3 21:51 phenomic.browser.ba0bbd0527652669ed85.js
-rw-r--r--   1 krister  staff    87K Apr  3 21:52 phenomic.browser.ba0bbd0527652669ed85.js.gz
```

After:
```
-rw-r--r--   1 krister  staff   301K Apr  3 21:54 phenomic.browser.ef41c4d45bc213bf0f62.js
-rw-r--r--   1 krister  staff    88K Apr  3 21:55 phenomic.browser.ef41c4d45bc213bf0f62.js.gz
```

### DIFF
--- a/themes/phenomic-theme-base/package.json
+++ b/themes/phenomic-theme-base/package.json
@@ -96,6 +96,7 @@
     "babel-preset-stage-2": "^6.22.0",
     "classnames": "^2.2.5",
     "css-loader": "^0.28.0",
+    "date-fns": "^1.28.2",
     "eslint": "^3.7.1",
     "eslint-loader": "^1.7.1",
     "eslint-plugin-react": "^6.4.0",
--- a/themes/phenomic-theme-base/src/components/PagePreview/index.js
+++ b/themes/phenomic-theme-base/src/components/PagePreview/index.js
@@ -1,4 +1,5 @@
 import React, { PropTypes } from "react"
+import parseDate from "date-fns/parse"
 import { Link } from "phenomic"
 
 import Button from "../../components/Button"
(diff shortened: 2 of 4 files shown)


MoOx:
Looks good to me. I guess this will close #944 poke @DavidWells @thangngoc89

MoOx:
Thanks for this, sorry for the lag :)

Issue #956, opened 2017-02-03
Modern static* => JAMstack?


bdougie:
hello @MoOx, has a quick question, sorry in advance if this is not that proper platform to discuss:

I see in the GitHub repo and on phenomic.io you point to static* and then describe. Any chance you would consider adding the term JAMstack instead or in addition of that.

JAMstack is a term to help describe what you are already describing. I realize as well that static doesn't quite cut it when describing what Phenomic and similar generators can do and I have noticed that as well, which is why I do a podcast on the very subject, [JAMstack Radio](http://www.heavybit.com/library/podcasts/jamstack-radio/), which I would love to have you on to chat about anytime in the future.

MoOx:
I could add a mention but this term is still new imo so using it in place of static will create probably more confusion.

bdougie:
You make a great point. I am hoping to make the term one that is more commonly used. Perhaps just adding a mention the in the description for static*: 😄 

---
All pages are available as HTML (works with no JavaScript) but users can get an enhanced navigation (when JavaScript is on). Phenomic provides a dynamic UX similar to apps, **aka JAMstack**. Try to navigate this website to see it by yourself.

I know it's still a stretch, but just trying to improve awareness for the term. As I mentioned, you are more than to come on the podcast to discuss the term `static*`. It is the exact conversation we are trying to have each week.

MoOx:
I will definitely integrate this new word in upcoming website update (after v1).

Meanwhile, feel free to add "aka JAMstack" with a link on the home.

MoOx:
Closed by #1000

PR #1000, opened 2017-04-05
adds reference to JAMstack on the homepage

bdougie:
Per #956 conversation.

quick question - What is the build command to view the output of the Homepage/index.js? I made this commit without review my changes locally

Running phenomic start or build just returns an error and after review your webpack config and npm scripts, I can only see the output. 

```
$ phenomic start
/assets doesn't exist or isn't a folder. Please check your 'assets' configuration. Note that if you don't need this option, you can set it up to `false`.
```

### DIFF
--- a/docs/src/layouts/Homepage/index.js
+++ b/docs/src/layouts/Homepage/index.js
@@ -35,6 +35,8 @@ const Homepage = (props) => (
         <h2>
           { "You can use it to build static" }
           <sup>{ "*" }</sup>
+          { "aka " }
+          <a href="https://jamstack.org/">{"JAMstack"}</a>
           { " websites." }
         </h2>
         <Link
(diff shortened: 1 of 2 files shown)


docs/src/layouts/Homepage/index.js
```diff
@@ -35,6 +35,8 @@ const Homepage = (props) => (
         <h2>
           { "You can use it to build static" }
           <sup>{ "*" }</sup>
+          { "aka " }
+          <a href="https://jamstack.org/">JAMstack)</a>
```
MoOx: should be `{ "JAMstack" }`

bdougie:
I fixed the missing `}` and also viewed it live. The link seemed odd as a purple link, so I made it white as well for the homepage. see it live here 👉🏾 http://biochemist-lookout-45604.netlify.com/

MoOx:
👌

PR #1001, opened 2017-04-06
Windows example on GitHub Pages page produces different behaviour

heedfull:
Fixed the Windows example so that it pushes to the remote gh_pages branch as per the macos/linux example.


MoOx:
Thanks!

PR #1004, opened 2017-04-10
Fixed (I think) broken link in this README

joostdecock:
While looking into adding reducers, I noticed the link in this README is broken.
I've updated it with what I think is the correct link, but please double-check.


MoOx:
This was due to a search and replace :)

Issue #779, opened 2016-09-24
HMR not being called on Windows


facelesspanda:
My browser is Chrome, stable version `53.0.2785.116 (Official Build) m (64-bit)` as of now.
OS: Windows 10, latest version (there were even updates this night... I hate the new auto-auto-updates D: )

I wanted to try out Phenomic, so I simply followed the setup guide.

went to a new empty folder, ran these in order and did nothing else:

```
npm i phenomic
```

```
./node_modules/.bin/phenomic setup
```

```
npm install
```

```
npm start
```

My browser page opened, with a stylized list of links to the docs and some posts that are in the boilerplate, so I thought "cool it works"

But then I started editing `content/index.md`, and I noticed nothing was output to the terminal, and no change was reflected in my browser — the dev console also didn't budge.

In small words: no HMR hot loading on windows 10 with the default config.
It must be an upstream issue, but still, it's cool to know here too.

facelesspanda:
Update:
It HMR reloads when my project is not in a path containing any type of brackets, or spaces.
It's a path issue somewhere upstream. Most likely chokidar.

thangngoc89:
Which page are you seeing the posts? Currently, hmr for markdown content only works on pages that contains 1 markdown posts, not the one with mutiple posts, which is filtered with enhanceCollection

facelesspanda:
I'm not sure I understand the question, but it was the boilerplate's homepage index.html file, at the root of the content tree. Though its layout contained a list of posts, its markdown content was updated correctly once my path was free of spaces or brackets. Editing the homepage index.js layout updated the page too. Didn't try editing the title of one of the posts from its own file.

thangngoc89:
I see. I guess this is something we can improve. I use Windows 10 too but I never name a file with spaces or brackets.

facelesspanda:
I have a folder structure where "categories" (like "development", "movies", "programs"...) are named in brackets to appear before any other folder when I sort by name. Never had problems before dealing with CLI tools, because a lot of gui software manage those by simply using the OS's file system management. Here, there are a lot of string manipulation directly between tools like Chokidar or Path, so it's easy to just not know who does what with there paths.

evenkiel:
The PR for #943 resolved the issue. Additionally, there was another solution posted by someone else on that same thread. Hot reloads worked fine for me for all md files after the fix. Please consider applying the fixes from the closed item to address this earlier opened issue.

FWIW, the path I was using had no spaces or special characters.

[...]

PR #1005, opened 2017-04-10
Fix HRM hot reload on windows (#779, #973)

### LINKED ISSUE #973: HMR error when editing markdown files requires a full reload after the first change

micnigh:
Normalize the `__filename` path of the phenomic loader to match webpack context filenames.

Applies fix proposed by @mktange in #974

Fixes #973, #779

### DIFF
--- a/src/loader/index.js
+++ b/src/loader/index.js
@@ -65,7 +65,7 @@ const loader = function(input: string) {
   const dataUrl = resourceUrl + "." + contentHash + ".json"
 
   const metadata = {
-    __filename: relativePath,
+    __filename: pathToUri(relativePath),
     __url: pathToUri("/", url),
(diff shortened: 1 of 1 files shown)


MoOx:
I will trust you on this :)

PR #1006, opened 2017-04-18
added www.codepreneur.io entry

codepreneur:
:)


codepreneur:
@MoOx ^

MoOx:
Won't be online quickly are we are freezing v0.x right now to focus on v1 (first public alpha at the end of this month)

PR #1007, opened 2017-04-19
Pass react-router location property to layouts

kennu:
I propose this change to pass on react-router's **location** property to the chosen layout component, so that the layout can access query string parameters and other location information as React properties.

Without this modification, I have had to use getChildContext in AppContainer to pass the location object to the layout, which doesn't work so well with the React property system. My goal is to make it easy to use componentWillReceiveProps() to react to query string changes.

### DIFF
--- a/src/components/PageContainer/component.js
+++ b/src/components/PageContainer/component.js
@@ -24,6 +24,7 @@ type Props = {
   getPage: Function,
   setPageNotFound: Function,
   logger: Object,
+  location: Object,
 }
 
 type Context = {
@@ -281,6 +282,7 @@ class PageContainer extends Component<DefaultProps, Props, void> {
           // head will be overwritten by "page"
           // (since page contains a head when loaded)
           head={ partialPageHead }
+          location={ props.location }
           { ...page }
         />
       )


MoOx:
Why don't you "just" use context in your layouts?

kennu:
There have been various problems and I may not remember all of them. But perhaps biggest one is that I can't see the previous context and react only when a query string parameter changes.

I.e. with properties I can do something like this:

    componentWillReceiveProps(newProps) {
      if (newProps.location.query.userId !== this.props.location.query.userId) {
        this.props.fetchUserProfile(newProps.location.query.userId)
      }
    }

This situation would occur when using a page like /user/ to show different content depending on the query string, and navigating e.g. from /user/?userId=1 to /user/?userId=2. When using context, I'm not sure if componentWillReceiveProps() is even called.

There may be other solutions, but I think that passing the location property from react router to layouts would be all in all quite helpful.

MoOx:
@kennu  we are freezing phenomic 0.x currently since our time and resources are limited for v1. You won't need this in v1 since API is more simple and different. Is this a heart breaker for you?

kennu:
Thanks for thinking about my heart. :-) I have solved this temporarily for myself by hotpatching the module with a single line to pass through the location object. I'll be quite happy to upgrade to 1.0 when the time is right.

PR #1022, opened 2017-05-13
Hot loading

MoOx:
Last commit is stupid since we don't use webpack-dev-server. But we still need a client to handle the HMR messages?

TODO after basic HMR ok via livereload: add react-hot-loader@next (3 beta) (but when HMR is ok, should not be a problem)

### DIFF
--- a/packages/phenomic-plugin-renderer-react/src/render/client.js
+++ b/packages/phenomic-plugin-renderer-react/src/render/client.js
@@ -9,6 +9,8 @@ import createStore from "../shared/store"
 
 const debug = require("debug")("phenomic:plugin:react")
 
+let store
+
 function render(routes: () => React$Element<any>) {
   debug("client rendering")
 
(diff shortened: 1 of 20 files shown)


MoOx:
I was tired. I just need to replace client with webpack-hot-middleware.

MoOx:
Ready for review.

packages/phenomic-plugin-renderer-react/src/render/client.js
```diff
@@ -18,11 +20,13 @@ function render(routes: () => React$Element<any>) {
   }
 
   const initialStateNode = document.getElementById("Hydration")
-  const store = createStore(
-    initialStateNode && initialStateNode.textContent
-      ? JSON.parse(initialStateNode.textContent)
-      : undefined,
-  )
+  store =
+    store ||
```
MoOx: This change is required for hot loading to work properly (as we re-render the full app). @bloodyowl can you confirm this won't cause any trouble for static rendering? (as it's in render/client.js ^^),

bloodyowl: not that beautiful but it should do without effects on SSR

PR #1024, opened 2017-05-22
Index content by arbitrary fields

bloodyowl:
Just kept the special case for `author` and `authors` for most users confort. might change later if we consider it too much.


MoOx:
Awesome!

PR #1031, opened 2017-05-25
chore(phenomic): avoid watcher for static build as it's completely useless

MoOx:
This might help to have CI ok on Windows.


MoOx:
Windows static build was probably not happy because of generation of "dist/*/index.html" file (as * is forbidden. So it's ok now as those are skipped (when cannot be resolved)

PR #1041, opened 2017-05-29
Add Zenika home page in the Showcase

Swiip:
Hi, I wanted to suggest zenika.com for the Showcase as it's realized with Phenomic now!

After using a home made build process using React & pre-rendering we migrate to Phenomic and it works like a charm. Thanks for this great tool.

Unfortunately, the source is closed but we didn't do much complex things apart using the existing Redux and integrate i18next. If anyone need tips on i18next, I'll be happy to share.


MoOx:
Gorgeous! Nice work. Will be on the website probably tonight ;)

Issue #1032, opened 2017-05-26
Windows static build don't generate page for collection


MoOx:
I guess the cause is here https://ci.appveyor.com/project/MoOx/phenomic/build/3580#L1998 (collection fetch return nothing but should return some items). Maybe a path thing...

bherila:
@MoOx can this be closed now?

MoOx:
Indeed

PR #1046, opened 2017-06-03
Fix windows path issue in file collector

shaunmarx:
Not sure if this is related to issue #1032 but no pages were returned as part of a collection for me while taking the alpha for a spin on my windows machine. This change fixes it.

### DIFF
--- a/packages/plugin-collector-files/src/index.js
+++ b/packages/plugin-collector-files/src/index.js
@@ -2,12 +2,16 @@ import path from "path";
 
 const debug = require("debug")("phenomic:plugin:collector-files");
 
+function normalizeWindowsPath(value: string) {
+  return value.replace(/(\/|\\)+/g, path.sep);
+}
+
 export function getKey(name: string, json: PhenomicTransformResult): string {
   if (json.data.path) {
     return json.data.path;
   }
   // normalize windows path
-  name = name.replace(/(\/|\\)+/g, path.sep);
+  name = normalizeWindowsPath(name);
   return path.join(
     path
       .dirname(name)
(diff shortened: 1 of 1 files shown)


MoOx:
Awesome. That's a good start!
I think there is still an issue but that's definitely making the situation better! Would have a deeper look later as I have to setup a VM to debug :(

shaunmarx:
No problem at all - let me know if there is any way I can help. Oh and great work on the alpha by the way.

MoOx:
Merging into alpha3 branch!

daine:
Hi MoOx and shaunmarx, this still doesn't work for me on Windows 10 even on alpha 3. 
The same codebase runs on my Mac, but not on my windows when getting the collection files. I get a `NotFoundError: Key not found in database`

shaunmarx:
@daine I believe this would have gone into the alpha 4 release so may be worthwhile to check using the latest. Just make sure you update the `@phenomic/preset-react-app` package to >= alpha.4 for the fix. 

I have been using alpha 4 on windows 10 for a while now and it seems to be working fine for me - at least `phenomic start` works fine. @MoOx mentioned that there may still be some issues on Windows, but I am unsure whether those related to paths specifically or not. The issues I have run into so far is:

* Db file locks hanging around which forces me to kill the node process every now and again
* phenomic build doesn't seem to work
 
I was going to look into the latter a bit today as I am getting close to releasing a new site and really need this working but just wanted to make sure it's not an issue on my side before creating anything. I am assuming it would be working on mac/linux though so may not be a blocker for me.

daine:
@shaunmarx Thanks for the info. I agree, build doesn't work for collections on Alpha 4 (both windows and unix/linux) so I'm hoping it'll be out on Alpha 5.

I know how you feel, I had built something last week and needed these parts working (they do on my mac) but this means other people (on windows) can't collaborate with me on them. Also, I had to revert to alpha 3 for the build to work.

MoOx:
I am adding a regression test that should block windows CI until we fix it. Stay tuned :D

shaunmarx:
Good to hear @MoOx ! Looking forward to seeing this resolved :)

PR #1053, opened 2017-06-04
Use `posts.node.list` instead of `posts.list`

ataylorme:
I was stepping through the getting started docs and noticed the conditional to display the list of posts has a bug.

`posts && posts.node && posts.list` should be `posts && posts.node && posts.node.list` as `list` is nested under `node`.

phenomic-getting-started-4-debugging


MoOx:
Thanks for reporting! Closed by c6544dafe25f41d83163c6645ebc93f4af598aff

PR #1054, opened 2017-06-04
Return a valid React element in the Layout definition

ataylorme:
When running through the docs example I was getting the error

> A valid React element (or null) must be returned. You may have returned undefined, an array or some other invalid object.

This was pretty easy to fix. I opted to sway the curly braces for parenthesis but you could also add a `return`.


MoOx:
Thanks again! Closed by e2ef6eeb8691c132b8395dfb0787b6b1a6883480

PR #1088, opened 2017-06-13
Add ced.photo to showcase

xuopled:


thangngoc89:
I love it. 

Just my 2 cents: the homepage doesn't have a title

xuopled:
Thank you for your feedback. I did not notice it. I will correct. Do you have a tool to verify SEO things?

MoOx:
Very nice work! Congrats!

PR #1115, opened 2017-07-15
Add reason example wip

bloodyowl:


packages/reason/package.json
```diff
@@ -22,7 +26,8 @@
   ],
   "description": "Phenomic Reason bindings",
   "dependencies": {
-    "reason-react": "^0.2.3"
+    "reason-react": "^0.2.3",
+    "@phenomic/plugin-renderer-react": "^1.0.0-alpha.4"
```
MoOx: Maybe we should make a preset? Because "What if no react"?

bloodyowl: I'll rename reason to reasonReact in case people want to use Tea or alternatives

MoOx:
CI fails because of flow. Just run "npm test" locally and you will get the issue :p

MoOx:
Ok found the issue. It's because of the flow config. We are currently mapping packages/*/lib to packages/*/src to get raw JS with types. Going to make a fix for reason.

MoOx:
@bloodyowl Warning: I adjusted/fixed this PR (with a rebase + force push), so be sure to pull before continuing on this :)

MoOx:
In short: flow does not like JS generated code by bucklescript, so we have to ignore all those stuff. Probably because buckescript does not use module.exports but exports directly.

MoOx:
@bloodyowl I rebased this branch, any plan to continue this effort?

bloodyowl:
@MoOx absolutely!

bloodyowl:
it works now @MoOx

MoOx:
Will have a look to test then. Thanks!

packages/reason/src/bodyRenderer.re
```diff
@@ -6,10 +6,17 @@ let make body::(body: PhenomicContent.jsBody) _children => {
...
       let props = ReactDOMRe.objToDOMProps originalProps;
-      ReactDOMRe.createElement
-        tag
-        ::props
-        [|ReasonReact.arrayToElement (Array.of_list (List.map renderChild reasonChildren))|]
+      switch tag {
+      | "a" =>
+        <Link toURL=[%bs.raw {| child[1].href |}]>
+          ([%bs.raw {|child[2]|}] |> Array.of_list |> ReasonReact.arrayToElement)
+        </Link>
```
MoOx: could we change this to use the thing I made in react BodyRenderer? The component I created handle properly external & relative url. It would be nice to be able to reuse it somehow as relying on rr Link is really not reliable in the real work (where markdown can give us any sort of link)

MoOx:
Going to merge even if that's still wip. Better than nothing :)

PR #1116, opened 2017-07-16
syntax error for jsx

YuyingWu:
There should be no semicolons in jsx, or there will be syntax errors. By the way, awesome tutorial, pretty clear, almost finish my blog after the guide :)

image


MoOx:
The fix needs to be done with a more complex change as this ; has been added by prettier-md.
I am going to handle this in #1111.
Thanks anyway!

PR #1117, opened 2017-07-16
replace `Layouts` with `PostLayouts`

YuyingWu:
`Layouts` is not defined among the context, but `PostLayouts` is. I guess it might be easier to understand while the layouts array has the same variable name, no matter `Layouts` or `PostLayouts`.


MoOx:
I have updated docs in #1111 and it's fixed there! Thanks!

PR #1123, opened 2017-07-21
react getting started guide fixes

soycode:
Running through the React getting started guide, hit what I think is a minor typo.


soycode:
Finished running through the tutorial and added a few other small corrections.

Some of the code is a bit ambiguous as to where to copy/paste it (in particular the pagination "Older posts" link, where the semicolon may or may not be correct depending where you put it), but dealing with that is arguably part of the learning experience, so my only suggestion there would maybe be a link to a general React resource.

MoOx:
Thanks for all those changes. I already got them fixed in a local WIP last month, sorry for not sharing those before. alpha 5 getting started has been reviewed and tested (and added as an example) so it should be ok. Thanks for your contribution and don't hesitate to make more PR like this, I will try to react more quickly :)

PR #1127, opened 2017-07-27
missing bracket

andrwh:


soycode:
FYI I have this fix along with a few other documentation fixes in #1123.

MoOx:
Thanks for this change. I already got it fixed in a local WIP last month, sorry for not sharing those before. alpha 5 getting started has been reviewed and tested (and added as an example) so it should be ok. Thanks for your contribution and don't hesitate to make more PR like this, I will try to react more quickly :)

PR #1128, opened 2017-07-31
Unicode spechars fix

revolunet:
fix for bad unicode characters in the markdown and `JSON.stringify`

i guess a similar fix should be applied in @1.0

see also : 

http://www.thespanner.co.uk/2011/07/25/the-json-specification-is-now-wrong/

https://bugs.chromium.org/p/v8/issues/detail?can=2&start=0&num=100&q=&colspec=ID%20Type%20Status%20Priority%20Owner%20Summary%20HW%20OS%20Area%20Stars&groupby=&sort=&id=1939

https://github.com/expressjs/express/issues/1132#issuecomment-10100419

### DIFF
--- a/src/_utils/serialize/index.js
+++ b/src/_utils/serialize/index.js
@@ -1,5 +1,5 @@
 // @flow
 
 export default (value: any): string => {
-  return JSON.stringify(value).replace(/\<\/script>/g, "<\\/script>")
+  return JSON.stringify(value).replace(/\<\/script>/g, "<\\/script>").replace(/\u2028/g,"\\u2028").replace(/\u2029/g,"\\u2029")
 }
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -128,7 +128,8 @@
     "stylelint": "^7.2.0",
     "stylelint-config-standard": "^16.0.0",
     "suppose": "^0.6.1",
-    "webpack": "^2.3.0"
+    "webpack": "^2.3.0",
+    "webpack-error-notification": "^0.1.7"
```
MoOx: Is this change really necessary? Not sure about this :)

revolunet:
ok, removed the `webpack-error-notification` dep. had an error with that at some point but whatever :)

MoOx:
0.21.2 is out. Hope ok as I didn't touch 0.x for a while and got issues when installing deps... 🙏

PR #1129, opened 2017-07-31
Removing ambigous words.

nicolas-zozol:
I've first created a app/App.js before getting a word from the compiler looking for ./App.js


nicolas-zozol:
Did dirty stuff late at night. I will make a wider scoped PR on documentation later after discussion on chat :)

PR #1146, opened 2017-08-21
minor update to the doc to make this work

ankitsinghaniyaz:


MoOx:
The getting started is in a single file on purpose (to keep thing simple) so I am not ok for the change in 3.md. Can you make a simpler PR (or adjust this one) for the first change? Thanks!

packages/preset-react-app/docs/getting-started/1.md
```diff
@@ -11,9 +11,11 @@ These lines will install Phenomic and its [react app preset](https://github.com/
 
 ⚠️ _If you don't have a ``package.json`` you can do this simple command to get
  a valid one:_
- 
+
 ```console
 echo '{"private":true}' > package.json
+# or you can also do
+yarn init
```
MoOx: I guess we should also specify `npm init` then? Just afraid this create more noize

ankitsinghaniyaz: I think instead of  `echo '{"private":true}' > package.json` we could just have `npm init` cause if the user installed `Phenomic` he is going to have npm anyways. We could remove the `yarn init` in that case.

MoOx: Like I said, I personnally don't like "npm init" because it's creating a huge file for nothing. So not sure it's necessary. It the echo command too weird?

ankitsinghaniyaz:
My bad on 3.md, from the code snippet I assumed that it was in a new file, so had some confusion! :sweat_smile:

packages/preset-react-app/docs/getting-started/1.md
ankitsinghaniyaz: You have a valid point and ya its just a preference. Then should I close the PR?

MoOx: Maybe yes. I think people know npm/yarn init anyway :)
But do not hesitate to make some others PRs if you feel the need, I like having discussions on docs!

MoOx:
@ankitsinghaniyaz do you think the tutorial will be more clear if it was split in several files? I was thinking having a single file will make it even easier to follow but I may be wrong :)

ankitsinghaniyaz:
@MoOx I think the idea of keeping it in the same file is good for getting started. If we were to move it to a new file then we may have to answer questions like where do we want to keep the files, so in the start, we can keep the choices minimum!
Also if we can just add a line about adding the content to the same file, so that someone skimming through the docs don't get confused. It may just be me doing too much react. :sweat_smile:

PR #1172, opened 2017-10-03
Fix borderRadius

Essential:
:lipstick: Minor UI improvement


MoOx:
Thanks for this PR. I will fix it by using a filter instead of a box-shadow :)

PR #1173, opened 2017-10-04
``@phenomic/core`` + ``@phenomic/api-client``: Make `path` parameter for query optional

MoOx:
Because we can.

### DIFF
--- a/packages/core/src/api/index.js
+++ b/packages/core/src/api/index.js
@@ -148,7 +148,7 @@ function createServer(db: PhenomicDB, plugins: PhenomicPlugins) {
   ) {
     debug(req.url, JSON.stringify(req.params));
     try {
-      const resource = await db.get(undefined, req.params["0"]);
+      const resource = await db.get(null, req.params["0"]);
       res.json(resource.value);
     } catch (error) {
       console.error(error);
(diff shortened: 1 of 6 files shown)


MoOx:
Note that I made this the quick and not really pretty way.

MoOx:
Smaller diff https://github.com/phenomic/phenomic/pull/1173/files?w=1

packages/core/src/api/index.js
```diff
@@ -142,6 +142,20 @@ function createServer(db: PhenomicDB, plugins: PhenomicPlugins) {
     }
   });
 
+  server.get("/item/*.json", async function(
+    req: express$Request,
+    res: express$Response
+  ) {
+    debug(req.url, JSON.stringify(req.params));
+    try {
+      const resource = await db.get(undefined, req.params["0"]);
```
bloodyowl: maybe should we use `null` to express that the absence of value is wished?

packages/plugin-collector-files/src/index.js
```diff
@@ -118,6 +118,10 @@ export default function() {
           debug(`collecting ${relativeKey} for path '${pathName}'`);
           return Promise.all([
             // full resource, not sorted
+            db.put(undefined, key, {
```
bloodyowl: same question about `null`

Issue #1153, opened 2017-09-06
Support asciidoc


kepinpin:
Please support asciidoc for templating: https://github.com/asciidoctor/asciidoctor.js

gandulio:
I would also love to see this!

ezralalonde:
I had hacked support for asciidoctor for the old Phenomic. It was kind of ugly, but not terribly difficult.

When I move to 1.0 I'll likely do it again. I'm also happy to help out if anybody else is taking this on.

ezralalonde:
This is enough for my needs:

- https://www.npmjs.com/package/phenomic-plugin-transform-asciidoc
- https://github.com/ezralalonde/phenomic-plugin-transform-asciidoc

Let me know if you're having problems or issues with the plugin.

Cheers,

MoOx:
Are you open to get this plugin into our repo? I can add you the write rights for the repo :)

ezralalonde:
I'm 100% open to contributing it to this repo. Thanks!

My task list for the asciidoc plugin to be "official" is roughly:

- [ ] write some tests for the plugin, however basic
- [ ] ensure `plugin-transform-asciidoc` can be used in place of `plugin-transform-markdown` or `plugin-transform-json` without any changes to default/example templates
- [ ] change internal source code style/naming to more closely follow conventions established in other transformers
- [ ] relicence to MIT

I think these are all quite easy. Let me know if there is anything else that comes to mind.

There are some other things that might be nice in the future—having plugins deal with source code in roughly the same manner, autolinking headers in the same way, being able to share CSS—that seem like much more work, and that can be dealt with a little later, if at all. Right now I think these would be a lot of effort for questionable returns. Expecting plugins to coöperate in this way is probably unrealistic (and very painful).

It will be a little while before I have a chance to do this work myself. If anybody else wants to take a try, feel free. I will submit a pull request to this issue when I have something.

Cheers!

PS. I also have very rough "how to write a plugin" notes that I hope to get together at some point, based on my initial experience writing a plugin. 
PPS. Again, it is a credit to everyone who works on Phenomic that I was able to develop this plugin. You are awesome!

PR #1174, opened 2017-10-06
Add Asciidoctor.js transformer plugin

ezralalonde:
I have a first attempt at the "official" Asciidoc plugin for #1153. Let me know if you have any comments or questions.

Hopefully I didn't break too much. 🔨 🏠 🔥 🚒  It works for me locally, but I've been fooled before.

This plugin more closely resembles the style (tests, README, etc.) of the existing 1.0 `markdown-transform-plugin` than of the other asciidoctor plugin I made previously. The main difference is that the new version uses `rehype` and plugins (so it has autolinking, etc. available), and highlights using `rehype-highlight` instead of inlining the DOM with `cheerio`/`lowlight`. It borrows the react/html output option from the markdown transformer, as well as the general structure.

There might still be some "nice to have" additions down the road, but this is a good start.

Summary of the changes:

+ stub flow interface for rehype-parse
+ stub flow interface for asciidoctor
+ transformer for `ad`, `adoc`, `asciidoctor` files using asciidoctor.js
+ simple jest tests
+ update yarn.lock dependencies

Thanks again, MoOx, bloodyowl, thangngoc89, and everyone else for sharing this awesome project! 🎉


ezralalonde:
Of course. Every developer's favourite things—timezones and windows paths.

I'll take another look in the next little while.

MoOx:
Looks like an awesome start! Thanks @ezralalonde!
I guess we should find a way to refactor code that looks redundant to what we have in the markdown plugins, and expose this somehow to make it even easier to write a transformer and share some logic.
Will merge as it is so we can start playing with it in next release and iterate!
Thanks again for the great work.

PR #1175, opened 2017-10-06
Add Link component to react-router so first post doesn't blow error

RVMendoza:
CURRENTLY: The tutorial has `BlogPost` component referring to a `Link` component. The tutorial has not instructed to add to the top import to `react-router`, so when the user runs the tutorial exactly as it says, it blows an error and doesn't load the first post as expected.

EXPECTED: The user can follow the tutorial exactly and get the result promised.

TO BE DONE: Add an extra line explaining that they need to update the react-router import to include link


MoOx:
Thanks!

PR #1183, opened 2017-10-13
pushing example with emotion-js styles

tgrecojs:
Please let me know if there's anything I can patch up 😄


MoOx:
That's an awesome start. Will merge as is an will make some minor tweaks and tests (as I like to automate minimal tests on examples to avoid regressions!).

Thanks a ton!

PR #1191, opened 2017-10-29
what is the use of phenomic .

Addictedk0126:


MoOx:
@Addictedk0126 what is the motivation for this PR? Did you wanted just to ask a question? If so feel free to use our chat https://github.com/MoOx/metalsmith-react/blob/master/README.md#is-it-possible-to-use-react-on-the-front-end-to-be-able-to-use-stuff-like-onclick

Issue #1165, opened 2017-09-25
Allow queries to be sorted by any fields


MoOx:
Currently "order" is restricted to id because it's leveldb default behavior.
To improve the situation, we will need to dig deeper into leveldb. Probably the thing called "iterator".

MoOx:
@bloodyowl mind taking a look to this issue?

luan-cestari:
Hi @MoOx , 

Does this issue require to use something like https://github.com/mirkokiefer/levelup-iterator  ? I saw https://github.com/phenomic/phenomic/blob/master/packages/core/src/db/index.js and it seems to have leveldown which have iterator but I can't see the usage and I thought leveldown would be mutual exclusive with levelup. 

Also about the issue, that means it would have a new arguments about the desire field to be used to sort in the entire API that use https://github.com/phenomic/phenomic/blob/master/packages/core/src/db/index.js ?

MoOx:
@luan-cestari I know that @bloodyowl attempted to take a look to this issue but I am not sure what was his conclusion (since he was the initial author of the current db+api).
From my search I saw indeed that iterators should be the solution. So I think a new field will be necessary for our API surface.

luan-cestari:
Is he working on this issue?

MoOx:
@bloodyowl will have a look and will get back to us asap.

bloodyowl:
I'm currently working on removing levelDB and using a plain JS-in-memory structure in order to free us from these limitations.
there should be some progress made by the end of the week, I'll let you both know.

MoOx:
Doable very easily since https://github.com/phenomic/phenomic/commit/bfe1aa84d0e6c079b6159cefda6341b67b4a299d (this commit add sorting in JS, default to date).

We will need to offer a public API for this.

PR #1193, opened 2017-11-01
Removed LevelDB

bloodyowl:
For now, the in-memoty layer just more or less mimics the level API.

Opens for more flexibility on how we let users query collections and items.


MoOx:
Can you check the diff we can see on circle-ci if it's normal https://circleci.com/gh/phenomic/phenomic/504 ?

MoOx:
This PR will be a breaking change, but I will implement #1165 with a default value to the date (so the release won't have any public breaking changes)

PR #1195, opened 2017-11-07
fix scroll behaviour when clicking on anchors

fazouane-marouane:
Hi,

We're having issues with internal links generated from markdown documents. Links to the same documents (normal html anchors) did not work as expected: when clicking on the links, no scrolling was done.

This was mainly due to a "event.preventDefault()". That's what this PR is all about.

### DIFF
--- a/packages/plugin-renderer-react/package.json
+++ b/packages/plugin-renderer-react/package.json
@@ -31,6 +31,8 @@
     "isomorphic-fetch": "^2.2.1",
     "prop-types": "^15.5.8",
     "react-hot-loader": "^3.0.0-beta.7",
+    "react-router-hash-link": "^1.1.1",
+    "recompose": "^0.26.0",
     "simple-json-fetch": "^1.0.1",
     "socket.io-client": "^1.7.2",
     "striptags": "^2.0.0",
--- a/packages/plugin-renderer-react/package.json
+++ b/packages/plugin-renderer-react/package.json
@@ -31,6 +31,7 @@
     "isomorphic-fetch": "^2.2.1",
     "prop-types": "^15.5.8",
     "react-hot-loader": "^3.0.0-beta.7",
+    "react-router-dom": "^4.2.2",
     "react-router-hash-link": "^1.1.1",
     "recompose": "^0.26.0",
     "simple-json-fetch": "^1.0.1",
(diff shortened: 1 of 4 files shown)


MoOx:
Sorry for the delay!

Shouldn't we check that it's the same page? What happen if a click on a link with an anchor to another page?

MoOx:
@fazouane-marouane up :)

fazouane-marouane:
Hi

I can try checking against the current location first. But the scroll probably won’t happen in the target page.
Sorry  I’ve got nothing better to propose.
If linking to outside pages with anchors is something that’s done usually, then maybe we need a more advanced Link component.
What are your thoughts on this?

fazouane-marouane:
@MoOx The proposed fix should be Okay now. I changed the Link component and I based it on `react-router-hash-link`.
Please, tell me what changes should I make to the PR mergeable.

packages/plugin-renderer-react/package.json
```diff
@@ -31,6 +31,9 @@
     "isomorphic-fetch": "^2.2.1",
     "prop-types": "^15.5.8",
     "react-hot-loader": "^3.0.0-beta.7",
+    "react-router-dom": "^4.2.2",
```
MoOx: This is not an option for us. We can't use RR v4, see https://github.com/phenomic/phenomic/issues/1130

fazouane-marouane: Well, finally someone came up with a solution not on phenomic side. you can find it here https://github.com/parisjs/parisjs-website/commit/a044ffb5250bc6f897df7fe4a8fdb54527315ed7

It consists of having a `hashLinkScroll` function:
```js
function hashLinkScroll() {
   const { hash } = window.location
 
   if (hash !== '') {
     setTimeout(() => {
       const id = hash.replace('#', '')
       const element = document.getElementById(id)
       if (element) element.scrollIntoView()
     }, 0)
   }
 }
```

Then, you'd use it like so `<Router ... onUpdate={ hashLinkScroll } > ...</Router>`.
So maybe, we just need to document this or find a way to automatically inject the `hashLinkScroll` in phenomic once and for all. What are your thoughts on this?

MoOx:
I guess documentation about his could be great. I will start to document the project in the coming weeks and will include this when I will add support for this. Thanks for digging into this problem!

Issue #1210, opened 2017-12-01
Better default webpack config


drazik:
Hello,

I just ran `npm run build` and saw that the resulting CSS file is not minified. So I looked at `@phenomic/plugin-bundler-webpack` [default webpack config file](https://github.com/phenomic/phenomic/blob/master/packages/plugin-bundler-webpack/src/webpack.config.js) and effectively, [css-loader's minimize option](https://github.com/webpack-contrib/css-loader#minimize) is not used at all.

Is this intentional or is it really missing ? If it is missing, I can propose a PR to fix it :)

MoOx:
In fact the default config is probably not good enough.
We should replace this one somehow by grabbing the one from CRA (via `react-scripts` https://github.com/facebookincubator/create-react-app/tree/9af042263f7c6e66b4c1402d2f6fc81250c6ce97/packages/react-scripts/config)
See #1190 and #1200 for more info.

MoOx:
PR welcome as it should not be that hard to get the config from 
`react-script` (via import/require) and merge it using spread operator and the things we need if relevant (hot loading probably?)

drazik:
Just to be sure that I understand what you want to achieve : the idea is to add `react-scripts` to `@phenomic/plugin-bundler-webpack` dependencies, then import its webpack configs (depending on the `NODE_ENV` and `PHENOMIC_ENV`) and maybe override some stuff ?

MoOx:
Exactly! Can feel weird in the first place, but for now it's for us the simplest way to rely on a good (and up to date) common config widely used (and so accepted)!
This might be revisited later, but for now, it's for us a good way to not spend time on creating a config to make everyone happy.

drazik:
OK I will try to work on that :) Thank you for the tips

MoOx:
If you need anything, just ask!

drazik:
I see that the project uses Lerna. To add a dependency on a particular package, do you use `lerna add module --scope=package` or directly `yarn add module` in the directory of the package ?

MoOx:
To be honest I don't use any of this solution. I add the dep by hand in the package.json of the relevant package.json and then I do `yarn` at the root of the repo :D

drazik:
Ok so I guess you don't really care how I install `react-scripts`

MoOx:
As you wish. Should be the same at the end anyway.

drazik:
I looked at it, began to work on a solution (using `webpack-merge`) and I have some questions : 

[...]

PR #1212, opened 2017-12-05
Better default webpack config

drazik:
Related to #1210. It is currently a not tested WIP. I just want to get some feedback from @MoOx for the moment. Does it match to what you imagined ?

Here are some random things : 

* To filter the rules, I used the `loader` property, not `test` as you suggested to me, because some of the `test`s are arrays, so I felt it was more difficult to use it in conditions
* I think some plugins used in `react-scripts` are not necessary in `phenomic` : `InterpolateHtmlPlugin`, `HtmlWebpackPlugin`, `DefinePlugin`... Do we keep them anyway ?

### DIFF
--- a/packages/plugin-bundler-webpack/package.json
+++ b/packages/plugin-bundler-webpack/package.json
@@ -33,7 +33,7 @@
     "find-cache-dir": "^0.1.1",
     "multili": "^1.1.0",
     "react-hot-loader": "^3.0.0-beta.7",
-    "react-scripts": "1.0.17",
+    "react-scripts": "^1.0.17",
     "source-map-support": "^0.4.0",
     "style-loader": "^0.16.1",
     "webpack": "^3.0.0",
--- a/packages/plugin-bundler-webpack/package.json
+++ b/packages/plugin-bundler-webpack/package.json
@@ -33,6 +33,7 @@
     "find-cache-dir": "^0.1.1",
     "multili": "^1.1.0",
     "react-hot-loader": "^3.0.0-beta.7",
+    "react-scripts": "1.0.17",
     "source-map-support": "^0.4.0",
     "style-loader": "^0.16.1",
     "webpack": "^3.0.0",
--- a/packages/plugin-bundler-webpack/src/webpack.config.js
+++ b/packages/plugin-bundler-webpack/src/webpack.config.js
@@ -3,22 +3,27 @@ import path from "path";
 import webpack from "webpack";
 import ExtractTextPlugin from "extract-text-webpack-plugin";
 
+const isStatic = process.env.PHENOMIC_ENV === "static";
+const isProduction = process.env.NODE_ENV === "production";
+
+const baseConfig = require("react-scripts/config/webpack.config." +
+  (isProduction ? "prod" : "dev"));
+
 module.exports = (config: PhenomicConfig) => ({
+  ...baseConfig,
   entry: {
     [config.bundleName]: [
-      process.env.PHENOMIC_ENV !== "static" &&
-        require.resolve("webpack-hot-middleware/client"),
-      process.env.PHENOMIC_ENV !== "static" &&
-        require.resolve("react-hot-loader/patch"),
+      !isStatic && require.resolve("webpack-hot-middleware/client"),
+      !isStatic && require.resolve("react-hot-loader/patch"),
       "./App.js"
     ].filter(item => item)
   },
   output: {
-    publicPath: "/", // @todo make this dynamic
+    ...baseConfig.output,
     path: path.isAbsolute(config.outdir)
       ? config.outdir
       : path.join(process.cwd(), config.outdir),
-    ...(process.env.PHENOMIC_ENV !== "static"
+    ...(!isStatic
       ? {
           filename: "phenomic/[name].js",
           chunkFilename: "phenomic/[name].chunk.js"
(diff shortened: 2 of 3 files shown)


packages/plugin-bundler-webpack/package.json
```diff
@@ -33,6 +33,7 @@
     "find-cache-dir": "^0.1.1",
     "multili": "^1.1.0",
     "react-hot-loader": "^3.0.0-beta.7",
+    "react-scripts": "1.0.17",
```
MoOx: Please use "^1.0.17"

packages/plugin-bundler-webpack/src/webpack.config.js
```diff
@@ -3,22 +3,27 @@ import path from "path";
 import webpack from "webpack";
 import ExtractTextPlugin from "extract-text-webpack-plugin";
 
+const isStatic = process.env.PHENOMIC_ENV === "static";
+const isProduction = process.env.NODE_ENV === "production";
```
MoOx: You cannot put this statically. Must be in the fonction to be safe (since we can adjust this during a function - we are not 100% sure that this file won't be imported (& so evaluated) before the code that will adjust this on runtime - FYI [here](https://github.com/phenomic/phenomic/blob/f03a7398322f4f8c56ef86353a11de5f1c4f435d/packages/core/src/commands/build.js#L122-L124) and [here](https://github.com/phenomic/phenomic/blob/f03a7398322f4f8c56ef86353a11de5f1c4f435d/packages/core/src/commands/start.js#L41-L43))

packages/plugin-bundler-webpack/src/webpack.config.js
```diff
@@ -3,22 +3,27 @@ import path from "path";
...
+
+const baseConfig = require("react-scripts/config/webpack.config." +
```
MoOx: Same as above, should be in the function.

MoOx:
About the unnecessary plugins, no big deal imo, we can improve that later. I think we should only remove what's really a problem. I would like to stick to CRA default as close as possible so people can just "drop" some existing code. Removing some plugin might make sense but I prefer to not do it unless necessary.

MoOx:
Any reason we have to maintain a specific css loader block? Is this because of the hmr?

drazik:
@MoOx I did the changes you requested.

I had an error because the template for `HtmlWebpackPlugin` didn't exist. We don't need to use it in Phenomic, so I deleted the plugin
 
After that, I saw that the css was written in `dist/static/css`. It was because CRA config declares a `cssFilename` that is used in `extractTextPluginOptions` variable to set the `publicPath` in the loader. I added `output.publicPath` in our config, but I had to copy/paste the loader configuration without it.

I also had to put a weird computed property with a ternary condition in the CSS loader because CRA is using `use` in dev config and `loader` in prod config.

Last thing, since we have to override the `ExtractTextPlugin` use by CRA, and they use it only in prod, I prefered to delete it in every case and just add a new one with our own config.

I just tested with `npm start` and `npm run build`. The config does a good job on my blog.

MoOx:
html plugin is supposed to be around https://github.com/facebookincubator/create-react-app/blob/4b55193f0ad479f26b0f5e153bb4b152a1ee03e7/packages/react-scripts/package.json#L45

Anyway, thanks for this, I will have a deeper look on some stuff before merging to try to reduce the maintenance even further. Thanks a lot for the work done. Stay tuned !

MoOx:
Could you fix the yarn lock issue? Rebase on master and just do "yarn", yarn will handle the conflict (it's a yarn feature!) then you can amend your commit :)
Are you interested to be added as a collaborator on the project so you can later directly work on branches?

drazik:
Yes the plugin is in the dependencies, but there is no HTML template file, so it crashed when it was trying to read it (CRA declares an `index.html`). Since we have nothing to do with such a file, I thought it was better to simply delete the plugin.

I will look the yarn lock issue tomorrow. And yeah, it would be so great to be added as a collaborator, thank you !

MoOx:
@drazik any luck with yarn lock conflict? (sorry I think that's the second time??)

bloodyowl:
@MoOx tkt je vais m'en occuper

MoOx:
I can still do it (merge in another branch, resolve & the merge into master). What I expect from you is more a code review what thoughts about the approach :)

packages/plugin-bundler-webpack/src/webpack.config.js
```diff
@@ -2,77 +2,128 @@ import path from "path";
...
-      : path.join(process.cwd(), config.outdir),
-    ...(process.env.PHENOMIC_ENV !== "static"
-      ? {
-          filename: "phenomic/[name].js",
-          chunkFilename: "phenomic/[name].chunk.js"
+module.exports = (config: PhenomicConfig) => {
+  const isStatic = process.env.PHENOMIC_ENV === "static";
+  const isProduction = process.env.NODE_ENV === "production";
+
+  const baseConfig = require("react-scripts/config/webpack.config." +
```
bloodyowl: I'm a bit worried to rely on a react repo for the webpack config (they are not coupled in Phenomic)

packages/plugin-bundler-webpack/src/webpack.config.js
```diff
@@ -2,77 +2,128 @@ import path from "path";
...
+    (isProduction ? "prod" : "dev"));
+
+  return {
+    ...baseConfig,
+    entry: {
+      [config.bundleName]: [
+        !isStatic && require.resolve("webpack-hot-middleware/client"),
+        !isStatic && require.resolve("react-hot-loader/patch"),
```
bloodyowl: I'd only add this if the React plugin is used

MoOx:
Totally agree. I was more like "until we have other renderer we could do this and fix later".
The correct approach would me more like that any plugin could send option to any other plugins (each plugins can accept options (in theory), we could add an api or something so plugin-renderer-react could pass a config to plugin-bundler-webpack somehow...

drazik:
I fixed the conflict in `yarn.lock`

MoOx:
It seems that tests are failing on Windows (see appveyor state). Can you take a look?

drazik:
Yes I will take a look. I'm pretty busy at the moment but I will take 10 minutes to do it this week ;)

MoOx:
@drazik would you like me to continue the effort here?

drazik:
Yes sorry I am very busy these days, and I don't know where the error on Windows come from. If you have time to investigate it's great :)

MoOx:
I will try to use another approach soon. @drazik thanks a lot for digging into this, that helped me to think about a better way to integrate CRA config!

drazik:
That's fine :) In the meantime, do you have some advices to deal with things the current webpack things doesn't handle ? I have some fonts to load, for example.

MoOx:
You can create your own custom config. Eg: https://github.com/MoOx/maxim-hylian.fr/blob/master/webpack.config.js

drazik:
Thank you :+1:

PR #1221, opened 2017-12-25
Link component accepts an object.toString as className

### LINKED ISSUE #1220: Link component: Accept an object as className
In redex.github.io app, [bs-css](https://github.com/SentiaAnalytics/bs-css) is used as css-in-js library. But the result is not a string but an object to `toString` method.

React is smart enough to call toString method on the result but `classnames` lib filters it out. 

image

thangngoc89:
This fixes #1220

### DIFF
--- a/packages/plugin-renderer-react/src/components/Link.js
+++ b/packages/plugin-renderer-react/src/components/Link.js
@@ -14,7 +14,7 @@ const origin = url =>
 type PropsType = {
   style?: Object,
   activeStyle?: Object,
-  className?: string,
+  className?: string | Object,
   activeClassName?: string,
   to?: string,
   href?: string,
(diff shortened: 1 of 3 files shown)


MoOx:
Why don't you just use cx before on your side?

thangngoc89:
After digging deep into this issue, I think that this is better to be resolved at upstream level rather than here:

1. React supports passing an object with toString method for props. It will calll the `toString` method itself. See test here https://github.com/facebook/react/blob/48616e591fe23c0b89b0823c3ec99bae2d7b6853/packages/react-dom/src/__tests__/DOMPropertyOperations-test.js#L63-L74

2. Since React supports is, classnames should support it too

3. I sent https://github.com/SentiaAnalytics/bs-css/issues/21 but it wasn't resolved.

So my solution (currently) is making my own Link binding (that use Link.js) to resolve it right now.

bloodyowl:
hum, I don't believe this should be fixed on our side

thangngoc89:
@bloodyowl yep. I think so too. Will try to fix upstream packages

PR #1230, opened 2018-02-23
Remove unneeded Js.Object.assign

ulrikstrid:
The Object.assign here seems to be unnecessary as we don't mutate or add more fields to it. We also doesn't save it for later use so we don't really care if it's mutated in JS-land either.

### DIFF
--- a/packages/api-client/src/phenomicApiClient.re
+++ b/packages/api-client/src/phenomicApiClient.re
@@ -1,5 +1,5 @@
-[@bs.module "@phenomic/api-client/lib/query"] external internalQuery : Js.t({..}) => Js.t({..}) =
-  "query";
+[@bs.module "@phenomic/api-client/lib/query"]
+external internalQuery : Js.t({..}) => Js.t({..}) = "query";
 
 type queryConfigItem = {
   path: string,
@@ -28,10 +28,15 @@ type queryConfig =
   | List(listConfig)
   | PaginatedList(paginatedListConfig);
 
-let query = (queryConfig) =>
+/* We need Js.Obj.assign here because OCaml can't handle returning multiple return types
+   from the same function. This makes it so it thinks it's just a Js.Obj.t */
+let query = queryConfig =>
   switch queryConfig {
   | Item(queryConfigItem) =>
-    Js.Obj.assign(Js.Obj.empty(), {"path": queryConfigItem.path, "id": queryConfigItem.id})
+    Js.Obj.assign(
+      Js.Obj.empty(),
+      {"path": queryConfigItem.path, "id": queryConfigItem.id}
+    )
   | List(queryConfigList) =>
     Js.Obj.assign(
       Js.Obj.empty(),


MoOx:
@bloodyowl any comment on why did you made this in the first place?

MoOx:
@ulrikstrid Tests are broken so maybe there is a reason it has been made this way :) (not sure yet)

ulrikstrid:
Seems to be a typing issue when returning plain objects as the switch returns different kinds of objects. I will try to work around it.

bloodyowl:
yeah, it was a nasty way to get the typing to match. optional keys should be marked as `Js.Nullable.t('a)`

MoOx:
@ulrikstrid do you think you can fix the all thing?

ulrikstrid:
There is no solution that is as clean as I would have wanted. Maybe just add a comment?

MoOx:
A comment lgtm.

ulrikstrid:
I can add a comment later today, or someone else can do it now.

It's pretty much the way it is because OCaml can't type multiple return types from the same function without wrapping it in another type.

bloodyowl:
I’ll do it tonight

ulrikstrid:
@MoOx @bloodyowl I added a comment, is it ok?

ulrikstrid:
Another solution would be this:
```ocaml
let query = queryConfig =>
  switch queryConfig {
  | Item(queryConfigItem) =>
      {
        "path": queryConfigItem.path,
        "id": Js.Undefined.return(queryConfigItem.id),
        "by": Js.Undefined.empty,
        "value": Js.Undefined.empty,
        "order": Js.Undefined.empty,
        "limit": Js.Undefined.empty,
        "after": Js.Undefined.empty
      }
  | List(queryConfigList) =>
      {
        "path": queryConfigList.path,
        "id": Js.Undefined.empty,
        "by": Js.Undefined.from_opt(queryConfigList.by),
        "value": Js.Undefined.from_opt(queryConfigList.value),
        "order": Js.Undefined.from_opt(queryConfigList.order),
        "limit": Js.Undefined.from_opt(queryConfigList.limit),
        "after": Js.Undefined.empty
      }
  | PaginatedList(queryConfigPaginatedList) =>
      {
        "path": queryConfigPaginatedList.path,
        "id": Js.Undefined.empty,
        "by": Js.Undefined.from_opt(queryConfigPaginatedList.by),
        "value": Js.Undefined.from_opt(queryConfigPaginatedList.value),
        "order": Js.Undefined.from_opt(queryConfigPaginatedList.order),
        "limit": Js.Undefined.from_opt(queryConfigPaginatedList.limit),
        "after": Js.Undefined.from_opt(queryConfigPaginatedList.after)
      }
  };
```
This will generate JS like this:
```js
        return {
                path: queryConfigItem[/* path */0],
                id: queryConfigItem[/* id */1],
                by: undefined,
                value: undefined,
                order: undefined,
                limit: undefined,
                after: undefined
              };
```
Would that be OK?

MoOx:
@ulrikstrid thanks for your input. @bloodyowl improved the situation in #1232. Are you ok with it?

bloodyowl:
we'll need to adapt the JS code a bit (as it relies on `hasOwnProperty` checks, but it'll be alright real soon)

ulrikstrid:
100% he beat me to it 👍

PR #1231, opened 2018-02-23
Remove unneeded div around bodyRenderer

ulrikstrid:
As discussed on gitter.

The diff is a bit noisy because of refmt, want me to revert and fix it?


MoOx:
@ulrikstrid can you check the diff if that's what you initially have in mind ? (I updated master with bs-platform 2.2)

ulrikstrid:
Yepp, exactly

MoOx:
👏

PR #1247, opened 2018-04-23
Create www.asimkt.com.md

asimqt:


MoOx:
Nice :)
You should add a visual indicator on the homepage that you should scroll to see the content below, I almost missed this part :)

asimqt:
@MoOx Oh thanks... Added to my to-do list

PR #1258, opened 2018-07-24
Fix: caching only http/https links

Kameshwaran:
The current source code has logic to catch all the href serverless links on <a> tags. It even mail addresses with mailto protocol. This commit adds a condition to catch only http/https link.


MoOx:
0.21.3 should be out. Tell me if that's ok!

PR #1271, opened 2019-01-28
chore: remove unused imports

DanielRuf:
This should fix at least one of the CI builds: https://github.com/phenomic/phenomic/issues/1270


DanielRuf:
So we have a few failing unit tests. Let's fix them after this PR.

MoOx:
I have fixed the remaining bug in master. Thanks for the pointer, I cut latest release in a train, wasn't the best idea I had!

PR #1279, opened 2019-04-30
[Feature] Update ReasonReact JSX to version 3

arecvlohe:
## Summary

This PR updates the ReasonReact jsx syntax to version 3, bsb-platform to 5.0.4, bs-react-helmet to 2.0.0, and reason-react to 0.7.0. In addition, the example reason-react-app is now updated to use jsx version 3.

### DIFF
--- a/examples/reason-react-app/App.js
+++ b/examples/reason-react-app/App.js
@@ -11,18 +11,12 @@ import { make as ErrorPage } from "./lib/es6/src/components/ErrorPage.bs.js";
 
 const routes = () => (
   <Router history={browserHistory}>
-    <Route
-      path="/"
-      component={withPhenomicApi(Home.jsComponent, Home.queries)}
-    />
+    <Route path="/" component={withPhenomicApi(Home.make, Home.queries)} />
     <Route
       path="/after/:after"
-      component={withPhenomicApi(Home.jsComponent, Home.queries)}
+      component={withPhenomicApi(Home.make, Home.queries)}
     />
-    <Route
-      path="blog/*"
-      component={withPhenomicApi(Post.jsComponent, Post.queries)}
-    />
+    <Route path="blog/*" component={withPhenomicApi(Post.make, Post.queries)} />
     <Route path="*" component={ErrorPage} />
     <Route path="404.html" component={ErrorPage} />
   </Router>
--- a/examples/reason-react-app/App.js
+++ b/examples/reason-react-app/App.js
@@ -7,7 +7,7 @@ import { withPhenomicApi } from "@phenomic/preset-react-app/lib/es6/src/phenomic
 
 import * as Home from "./lib/es6/src/components/Home.bs.js";
 import * as Post from "./lib/es6/src/components/Post.bs.js";
-import ErrorPage from "./lib/es6/src/components/ErrorPage.bs.js";
+import { make as ErrorPage } from "./lib/es6/src/components/ErrorPage.bs.js";
 
 const routes = () => (
   <Router history={browserHistory}>
--- a/examples/reason-react-app/App.js
+++ b/examples/reason-react-app/App.js
@@ -7,7 +7,7 @@ import { withPhenomicApi } from "@phenomic/preset-react-app/lib/es6/src/phenomic
 
 import * as Home from "./lib/es6/src/components/Home.bs.js";
 import * as Post from "./lib/es6/src/components/Post.bs.js";
-import { make as ErrorPage } from "./lib/es6/src/components/ErrorPage.bs.js";
+import ErrorPage from "./lib/es6/src/components/ErrorPage.bs.js";
 
 const routes = () => (
   <Router history={browserHistory}>
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "babel-core": "7.0.0-bridge.0",
     "babel-eslint": "^10.0.0",
     "babel-jest": "^23.6.0",
-    "bs-platform": "^5.0.0",
+    "bs-platform": "^5.0.3",
     "chalk": "^1.1.3",
     "cross-env": "^2.0.0",
     "cross-spawn": "^5.1.0",
@@ -147,5 +147,8 @@
     "website",
     "examples/*",
     "packages/*"
-  ]
+  ],
+  "dependencies": {
+    "reason-react": "^0.7.0"
+  }
 }
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "babel-core": "7.0.0-bridge.0",
     "babel-eslint": "^10.0.0",
     "babel-jest": "^23.6.0",
-    "bs-platform": "^5.0.3",
+    "bs-platform": "^5.0.4",
     "chalk": "^1.1.3",
     "cross-env": "^2.0.0",
     "cross-spawn": "^5.1.0",
@@ -147,8 +147,5 @@
     "website",
     "examples/*",
     "packages/*"
-  ],
-  "dependencies": {
-    "reason-react": "^0.7.0"
-  }
+  ]
 }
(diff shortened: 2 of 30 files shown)


packages/plugin-renderer-react/src/link.re
```diff
@@ -1,23 +1,13 @@
...
-    ~props={
-      "href": Js.Nullable.fromOption(href),
-      "style": Js.Nullable.fromOption(style),
-      "activeStyle": Js.Nullable.fromOption(activeStyle),
-      "className": Js.Nullable.fromOption(className),
-      "activeClassName": Js.Nullable.fromOption(activeClassName),
-    },
-    children,
-  );
+[@react.component]
```
arecvlohe: I should probably revert this as this is not a `[@react.component]`

packages/plugin-renderer-react/src/phenomicRendererReactClient.re
```diff
@@ -1,16 +1,16 @@
 [@bs.module "@phenomic/plugin-renderer-react/lib/client"]
 external createContainer_:
-  (ReasonReact.reactClass, Js.t({..})) => ReasonReact.reactClass =
+  (ReasonReact.reactClass, Js.t({..})) => React.element =
```
arecvlohe: Neither are these.

yarn.lock
```diff
@@ -2982,6 +2982,11 @@ bs-platform@^5.0.0:
   resolved "https://registry.yarnpkg.com/bs-platform/-/bs-platform-5.0.0.tgz#f9fded818bafc3891aeb85eb4e0d95b8d8f8b4d7"
   integrity sha512-zxLobdIaf/r7go47hOgxcd6C0ANh7MYMEtZNb9Al7JdoekzFZI50F4GpZpP8kMh9Z+M5PoPE1WuryT4S8mT8Kg==
 
+bs-platform@^5.0.3:
```
arecvlohe: This is duped.

package.json
```diff
@@ -147,5 +147,8 @@
     "website",
     "examples/*",
     "packages/*"
-  ]
+  ],
+  "dependencies": {
+    "reason-react": "^0.7.0"
+  }
```
MoOx: This isn't necessary.
You should add/modify this dep in appropriate package & run `yarn` at the root of the repo. Yarn _workspace_ feature handle all the necessary work to correctly install deps & make symlink if necessary.

package.json
```diff
@@ -22,7 +22,7 @@
     "babel-core": "7.0.0-bridge.0",
     "babel-eslint": "^10.0.0",
     "babel-jest": "^23.6.0",
-    "bs-platform": "^5.0.0",
+    "bs-platform": "^5.0.3",
```
MoOx: Please update this deps in example/reason-react-app & in packages/bs-platform too, for consitency :)

packages/plugin-renderer-react/src/link.re
MoOx: This is a component, so this looks good.

packages/plugin-renderer-react/src/phenomicRendererReactClient.re
MoOx: Here indeed that's different this function take a component & return a component. So `React.component('a)` should be used instead reactClass

MoOx:
To be sure the thing is working, you also need to edit example/reason-react-app.
That's where you can specify an actual reason-react version & run a build :)

Good start anyway!

arecvlohe:
Cool! I will try and work on this more tonight, but might be tomorrow. Also, I realized that we might have to use `ReasonReactCompat.wrapReasonReactForReact` in cases where we can't convert jsx 2 to jsx 3.

arecvlohe:
@MoOx Still just working along and making necessary changes but running into an issue that I am not able to diagnose:

```shell
sh: /Users/arecvlohe/Documents/development/phenomic/node_modules/bs-platform/lib/bsppx.exe -bs-jsx 3: No such file or directory
File "", line 1:
Error: Error while running external preprocessor
Command line: '/Users/arecvlohe/Documents/development/phenomic/node_modules/bs-platform/lib/bsppx.exe -bs-jsx 3' '/var/folders/h6/x5vgndfs0dv_y2017t23x51m0000gn/T/camlppx3bf7df' '/var/folders/h6/x5vgndfs0dv_y2017t23x51m0000gn/T/camlppxf075b9'
```

The error will show in your editor at the very top of `link.re`. It seems like everything else builds fine. I was having this issue in other files but it's only an issue in one now. I will take a look at this some more tomorrow night but wanted to leave it here in case you could take a look at it.

packages/plugin-renderer-react/src/phenomicRendererReactClient.re
```diff
@@ -1,16 +1,14 @@
 [@bs.module "@phenomic/plugin-renderer-react/lib/client"]
-external createContainer_:
-  (ReasonReact.reactClass, Js.t({..})) => ReasonReact.reactClass =
+external createContainer_: (React.component('a), Js.t({..})) => React.element =
```
MoOx: I think it's more returning the same React.component (kind of), & same below. Right?

arecvlohe: What do you mean here?

arecvlohe:
@MoOx Thanks for the context, it looks like the error is coming from RLS. I still have errors when running `yarn run test` so I will into that more today.

arecvlohe:
@MoOx Leaving a comment here because I can't figure this out. When I run `yarn run transpile` I am getting a syntax error in `link.re` on the line where it has `React.element`. I am not sure why this is happening. I Also found it strange that in `Link.js` no children are passed but when it's used a `React.array` is passed to `<Link />`. Maybe this is shows a lack of understanding but wanted to make note of it so I can learn something. Anyhow, the syntax error is main thing that is blocking me so any help would be appreciated!

arecvlohe:
@MoOx Sorry for spamming you on a Saturday but I think the code is now ready. After running `yarn run test`, all the tests pass. It looks as though there is a build error though, something related to using `&&`. You probably would know more about that than me. Anyhow, have a good weekend and let me know if the code needs any more tweaking!

examples/reason-react-app/App.js
```diff
@@ -7,22 +7,16 @@ import { withPhenomicApi } from "@phenomic/preset-react-app/lib/es6/src/phenomic
 
 import * as Home from "./lib/es6/src/components/Home.bs.js";
 import * as Post from "./lib/es6/src/components/Post.bs.js";
-import ErrorPage from "./lib/es6/src/components/ErrorPage.bs.js";
+import { make as ErrorPage } from "./lib/es6/src/components/ErrorPage.bs.js";
```
MoOx: Maybe we could avoid that by doing `let default = make`?

arecvlohe: Yeah, that makes sense. I don't have a preference either way.

arecvlohe:
@MoOx Done. I can see the benefit of splitting it up into different packages. Might make it easier to maintain overall. If more people contribute/maintain then it would be easier for one person to maintain one library instead of many.

packages/plugin-renderer-react/src/phenomicRendererReactClient.re
MoOx: Should this be `(React.component('a), Js.t({..})) => React.component('a)` ? an hoc isn't returning a rendered element, but almost the same component

arecvlohe:
@MoOx Done. I thought this made more sense `(React.component('a), Js.t({..})) => React.component('b)` since it's a HOC, it should be going from '`a` to `'b`.

arecvlohe:
@MoOx ping

arecvlohe:
@MoOx Ping

MoOx:
1.0.0 is out on npm (just include this change). Thanks for helping!

arecvlohe:
@MoOx Awesome!

Issue #1280, opened 2019-08-25
Page 3 of tutorial is not correct


lukos:
In the getting started tutorial, section 3, the code listing is incorrect and shows `BlogPostContainer` declared after `createApp`, which is where it is used. The listing should have `BlogPostContainer`  declared first. There is not even an error when this happens, it just silently fails to display the md content.

Secondly, the explanation that starts with "In our example above, we are doing a query on the path" does not read correctly. The query is actually _content/posts_, not posts. Also, when it talks about id being the filename, it incorrectly states "in our case posts/first-post" but the actual id from the path would be "first-post", not "posts/first-post".

MoOx:
@lukos Sorry about the issue. You can refer to https://github.com/phenomic/phenomic/tree/master/examples/react-app-getting-started which should work.
PR to fix doc are welcome :)

PR #1283, opened 2019-09-20
#1280 Fix explanation in tutorial page

lukos:
Fix incorrect explanation and move code to the position where it will work.


MoOx:
Thanks!

